PR #1, opened 2021-05-04
CFE-3395: promise module for git

tranchitella:
Changelog: None
Ticket: CFE-3395

Signed-off-by: Fabio Tranchitella <fabio.tranchitella@northern.tech>


promise_types/git/git.py
```diff
@@ -1,40 +1,180 @@
...
-            if name != "repo":
-                raise ValidationError(f"Unknown attribute '{name}' for git promises")
-            if name == "repo" and type(value) is not str:
-                raise ValidationError(f"'repo' must be string for git promise types")
-
-    def evaluate_promise(self, promiser, attributes):
-        folder = promiser
-        url = attributes["repo"]
+    def validate_promise(self, promiser: str, attributes: Dict):
+        attributes.setdefault("dest", promiser)
```
olehermanse, line 50: What is the advantage of being able to set a different promiser and supply `dest` as an attribute?

promise_types/git/git.py
```diff
@@ -1,40 +1,180 @@
...
+        try:
+            GitPromiseTypeModel(**attributes)
+        except PydanticValidationError as e:
+            errors = [
+                ".".join(map(str, err["loc"])) + ": " + err["msg"] for err in e.errors()
+            ]
+            raise ValidationError(", ".join(errors))
```
olehermanse, line 57: Cool!

promise_types/git/git.py
```diff
@@ -1,40 +1,180 @@
...
+                except subprocess.CalledProcessError as e:
+                    error = e.output.decode()
+                    self.log_error(f"Failed fetch: " + error)
+                    return (Result.NOT_KEPT, [f"{safe_promiser}_update_failed"])
+
+        # everything okay
+        return (Result.KEPT, classes)
+
+    def _git_command(self, model: GitPromiseTypeModel):
+        return model.executable or "git"
```
olehermanse, line 169: Is this necessary? `model.executable` already defaults to `"git"`, right?

promise_types/git/git.py
```diff
@@ -1,40 +1,180 @@
...
+                        [self._git_command(model), "rev-parse", "--abbrev-ref", "HEAD"],
+                        cwd=model.dest,
+                        env=self._git_envvars(model),
+                    )
+                    if output.decode("utf-8").strip() != model.version:
+                        output = subprocess.check_output(
+                            [self._git_command(model), "checkout", model.version],
+                            cwd=model.dest,
+                            env=self._git_envvars(model),
+                        )
```
olehermanse, line 148: I think we should also check the commit SHA of HEAD, and support supplying commit SHA as `version`

olehermanse (approved):
Thanks for your work on this! I am okay with merging as-is, it is a big improvement over the example we had. I left some comments, and I will also add some more high level thoughts to [the ticket](https://tracker.mender.io/browse/CFE-3395).

promise_types/git/git.py
tranchitella: the (sanitized) promiser is used to generate classes (_path_to_my_repo_updated), which are not the easiest thing to read and write on Earth. As a user, I would prefer having `my_repo_updated` and being able to specify a destination path that won't influence the name of the class. At least, this was my naive thinking. :)

promise_types/git/git.py
tranchitella: As I mentioned in the ticket, this stuff should work out-of-the-box and provided by cfengine.py

promise_types/git/git.py
tranchitella: no, it is not necessary, I'll rework it as a follow-up PR

promise_types/git/git.py
tranchitella: I'll rework it as a follow-up PR

PR #2, opened 2021-05-05
Git promise module improvements

tranchitella:


tranchitella:
@olehermanse FYI

promise_types/git/git.py
```diff
@@ -151,23 +159,28 @@ def evaluate_promise(self, promiser: str, attributes: Dict):
...
                             [
                                 model.executable,
-                                "merge",
-                                model.remote + "/" + model.version,
+                                "diff",
+                                f"..{model.remote}/{model.version}",
                             ],
                             cwd=model.dest,
                             env=self._git_envvars(model),
                         )
```
olehermanse: At this point I was a little confused as to why detached means skip the diff step, but I guess it's becaus if `detached == True`, we have already checked out a commit (SHA).

olehermanse (approved):
Looks great, thanks!

PR #3, opened 2021-05-06
CFE-3396: promise module for systemd

tranchitella:
Changelog: none
Ticket: CFE-3396

Signed-off-by: Fabio Tranchitella <fabio.tranchitella@northern.tech>


tranchitella:
@olehermanse ping

promise_types/systemd/README.md
```diff
@@ -0,0 +1,108 @@
...
+      depth => "1",
+      executable => "/bin/git",
+      force => "true",
+      recursive => "true",
+      remote => "origin",
+      ssh_options => "UserKnownHostsFile=/dev/null",
+      update => "true",
+      version => "master";
+}
+```
```
olehermanse, line 92: Leftovers. @tranchitella My favorite is Carbonara 🍝 🧀

tranchitella: @olehermanse yep, I figured it out and force-pushed before seeing this comment.

tranchitella: @olehermanse when you come to visit us in Trieste, I can prepare the best Carbonara you ever ate.

olehermanse:
@nickanderson please review the policy examples and test.

promise_types/systemd/systemd.py
```diff
@@ -0,0 +1,389 @@
...
+                    "show",
+                    model.name,
+                    "-p",
+                    "ActiveState",
+                    "-p",
+                    "SubState",
+                    "-p",
+                    "UnitFileState",
+                ]
+            )
```
olehermanse, line 122: Not the right line number, but I noticed something during your demo:

1. Some stderr output of CLI was printed directly (from git or systemctl, I don't remember)
   a. Should be captured and printed as log level verbose, or log level error if there was an error
2. I think you had 3 repeated log messages which said the same thing ("Updating systemd service" or something similar)
   a. Was this because the promise was evaluated and resulted in repairs 3 times? Normally, it should be REPAIRED with 1 INFO message, then KEPT with 0 INFO messages

promise_types/systemd/README.md
```diff
@@ -0,0 +1,114 @@
...
+
+## Synopsis
+
+* *Name*: `systemd`
+* *Version*: `0.1.0`
+* *Description*: Create and manage services using systemd
+
+## Requirements
+
+* A system managed by systemd.
```
vpodzime, line 11: Should this mention the minimum version of Python required? Or are those different requirements @olehermanse ?

promise_types/systemd/README.md
```diff
@@ -0,0 +1,114 @@
...
+| `masked` | `boolean` | Mask the service, making it impossible to start | No | `false` |
+| `name` | `string` | The name of the service | Yes | Promiser |
+| `replace` | `boolean` | Replace service unit file if it already exists | No | `False` |
+| `state` | `string` | State of the service: `started`, `stopped` and `removed` are idempotent, `restart` and `reload` are not. | Yes | - |
+| `unit_description` | `string` | Free-form string describing the unit | No | - |
+| `unit_requires` | `slist` | Units which must be started along with the service | No | - |
+| `unit_wants` | `slist` | Units which should be started along with the service | No | - |
+| `unit_after` | `slist` | Units which must be started after the service | No | - |
+| `unit_before` | `slist` | Units which must be started before the service | No | - |
+| `unit_extra` | `string` | Multiline string to be appended to the `Unit` section of the service file | No | - |
```
vpodzime, line 29: I'd prefer this to be an `slist` of lines.

promise_types/systemd/README.md
```diff
@@ -0,0 +1,114 @@
...
+| `service_user` | `string` | User under which the service should be executed | No | - |
+| `service_group` | `string` | Group under which the service should be executed | No | - |
+| `service_nice` | `integer` | Default nice level for the service | No | - |
+| `service_oom_score_adjust` | `integer` | Adjustment level for the Out-Of-Memory killer | No | - |
+| `service_exec_start` | `string` | Commands that are executed when this service is started | No | - |
+| `service_exec_start_pre` | `string` | Commands that are executed before `exec_start` commands | No | - |
+| `service_exec_start_post` | `string` | Commands that are executed after `exec_start` commands | No | - |
+| `service_exec_stop` | `string` | Commands to execute to stop the service started via `exec_start` | No | - |
+| `service_exec_stop_post` | `string` | Commands that are executed after the service is stopped | No | - |
+| `service_exec_reload` | `string` | Commands to execute to trigger a configuration reload in the service | No | - |
```
vpodzime, line 41: "commands" -- how should they be separated? I think `slist` would be a better option for the `service_exec_` attributes.

promise_types/systemd/README.md
```diff
@@ -0,0 +1,114 @@
...
+| `service_restart` | `string` | When the service must be restarted | No | - |
+| `service_restart_sec` | `string` | Time to sleep before restarting the service | No | - |
+| `service_timeout_sec` | `string` | Maximum waiting time for start/stop commands processing | No | - |
+| `service_environment` | `map` | Environment variables | No | - |
```
vpodzime, line 45: `map`?

promise_types/systemd/README.md
```diff
@@ -0,0 +1,114 @@
...
+| `service_environment_file` | `string` | Path to a file containing the environment variables | No | - |
+| `service_working_directory` | `string` | Working directory | No | - |
+| `service_standard_input` | `string` | Path to the standard input file descriptor | No | - |
+| `service_standard_output` | `string` | Path to the standard output file descriptor | No | - |
+| `service_standard_error` | `string` | Path to the standard error file descriptor | No | - |
+| `service_tty_path` | `string` | Path to the tty if required by the `standard_*` options | No | - |
+| `service_extra` | `string` | Multiline string to be appended to the `Service` section of the service file | No | - |
+| `install_wanted_by` | `slist` | Units which are wanted by the service | No | - |
+| `install_required_by` | `slist` | Units which are required by the service | No | - |
+| `install_extra` | `string` | Multiline string to be appended to the `Install` section of the service file | No | - |
```
vpodzime, line 55: `slist` of lines.

promise_types/systemd/README.md
```diff
@@ -0,0 +1,114 @@
...
+
+## Examples
+
+Start a service named `myservice`:
```
vpodzime, line 59: I think this should use the "idempotent wording", so something like "Make sure service named `myservice` is started."

vpodzime: And the same applies to the examples below.

promise_types/systemd/systemd.py
```diff
@@ -0,0 +1,389 @@
...
+    BaseModel,
+    ValidationError as PydanticValidationError,
+    validator,
+)
+
+
+SYSTEMD_LIB_PATH = "/lib/systemd/system"
+
+
+class SystemdPromiseTypeStates:
```
vpodzime, line 18: This could be an `Enum`.

promise_types/systemd/systemd.py
```diff
@@ -0,0 +1,389 @@
...
+
+    def validate_promise(self, promiser: str, attributes: Dict):
+        attributes.setdefault("name", promiser)
+        if type(attributes.get("environment")) == str:
+            attributes["environment"] = json.loads(attributes["environment"])
+        try:
+            SystemdPromiseTypeModel(**attributes)
+        except PydanticValidationError as e:
+            errors = [
+                ".".join(map(str, err["loc"])) + ": " + err["msg"] for err in e.errors()
```
vpodzime, line 94: I'd put `for err in e.errors()` on a separate line here to improve readability.

promise_types/systemd/systemd.py
```diff
@@ -0,0 +1,389 @@
...
+                    "-p",
+                    "ActiveState",
+                    "-p",
+                    "SubState",
+                    "-p",
+                    "UnitFileState",
+                ]
+            )
+            service_status = dict(
+                k.split("=", 1) for k in output.decode("utf-8").strip().split("\n")
```
vpodzime, line 124: `splitlines()` can be used instead of `split("\n")`

promise_types/systemd/systemd.py
```diff
@@ -0,0 +1,389 @@
...
+            attributes["environment"] = json.loads(attributes["environment"])
+        try:
+            SystemdPromiseTypeModel(**attributes)
+        except PydanticValidationError as e:
+            errors = [
+                ".".join(map(str, err["loc"])) + ": " + err["msg"] for err in e.errors()
+            ]
+            raise ValidationError(", ".join(errors))
+
+    def evaluate_promise(self, promiser: str, attributes: Dict) -> (str, List[str]):
```
vpodzime, line 98: Please split this into multiple (private) methods based on the request.

promise_types/systemd/systemd.py
```diff
@@ -0,0 +1,389 @@
...
+
+        # remove the service
+        if model.state == SystemdPromiseTypeStates.REMOVED:
+            # stop the service, in case it is running
+            if (
+                service_status["ActiveState"] == "active"
+                and service_status["SubState"] == "running"
+            ):
+                try:
+                    self.log_error(f"Stopping the service {model.name}")
```
vpodzime, line 139: `log_info`

vpodzime: The `info:` messages should inform about changes that happened. So "Stopped the service..." and only in case it actually happens. Things like "Stopping the service..." should be `verbose:`

promise_types/systemd/systemd.py
```diff
@@ -0,0 +1,389 @@
...
+        # render the template of the service
+        service_template = self._render_service_template(model)
+
+        # create the service file if it doesn't exist, or replace it if the content
+        # doesn't match our template and replace is true
+        service_path = os.path.join(SYSTEMD_LIB_PATH, f"{model.name}.service")
+        if (
+            not os.path.exists(service_path)
+            or model.replace
+            and open(service_path).read() != service_template
```
vpodzime, line 188: The file can fail to open because of permissions.

promise_types/systemd/systemd.py
```diff
@@ -0,0 +1,389 @@
...
+        ):
+            self.log_info(f"Installing the service {model.name}")
+            try:
+                open(service_path, "w").write(service_template)
```
vpodzime, line 192: Are files created and used this way correctly closed? It didn't use to be the case, AFAICT, `with open() as f:` was needed to make sure the context manager takes care of closing the file.

promise_types/systemd/systemd.py
```diff
@@ -0,0 +1,389 @@
...
+        if model.daemon_reexec:
+            self.log_info("Running daemon-reexec")
+            try:
+                subprocess.check_output(["systemctl", "daemon-reexec"])
+            except subprocess.CalledProcessError as e:
+                error = e.output.decode()
+                self.log_error(f"Failed to run systemctl daemon-reexec: {error}")
+                return (Result.NOT_KEPT, [f"{safe_promiser}_daemon_reexec_failed"])
+            classes.append(f"{safe_promiser}_daemon_reexec")
+
```
vpodzime, line 209: Please remove this blank line to keep the `if` and `elif` one right after the other one. Otherwise it easily gets lost in the series of `if`s.

promise_types/systemd/systemd.py
```diff
@@ -0,0 +1,389 @@
...
+            ("service", "service_environment_file", "EnvironmentFile"),
+            ("service", "service_working_directory", "WorkingDirectory"),
+            ("service", "service_standard_input", "StandardInput"),
+            ("service", "service_standard_output", "StandardOutput"),
+            ("service", "service_standard_error", "StandardError"),
+            ("service", "service_tty_path", "TTYPath"),
+            ("install", "install_wanted_by", "WantedBy"),
+            ("install", "install_required_by", "RequiredBy"),
+        ):
+            value = getattr(model, attr, None)
```
vpodzime, line 361: `, None` is unnecessary here, I think.

promise_types/systemd/systemd.py
```diff
@@ -0,0 +1,389 @@
...
+        #
+        if model.service_extra:
+            blocks["unit"].append(model.service_extra)
+        service_section = "\n".join(blocks["service"])
+        #
+        if model.install_extra:
+            blocks["unit"].append(model.install_extra)
+        install_section = "\n".join(blocks["install"])
+        #
+        return f"[Unit]\n{unit_section}\n\n[Service]\n{service_section}\n\n[Install]\n{install_section}\n"
```
vpodzime, line 385: There's a general rule to include something like `# Rendered by CFEngine blah, blah, blah. Do not edit directly` to rendered files.

promise_types/systemd/systemd.py
```diff
@@ -0,0 +1,389 @@
...
+
+
+if __name__ == "__main__":
+    SystemdPromiseTypeModule(record_file_path="/tmp/test.log").start()
```
vpodzime, line 389: Nice!

promise_types/systemd/test.cf
```diff
@@ -0,0 +1,74 @@
...
+        scope => "namespace";
+}
+
+#######################################################
+
+bundle agent check
+{
+  classes:
+      "file_ok"
+        if => fileexists("/lib/systemd/system/sample.service");
```
vpodzime, line 50: This should also compare the content of the file with some expected content.

vpodzime (commented):
Looks good to me otherwise. Please keep in mind that some of the comments below apply to multiple places in the code.

promise_types/systemd/test.cf
```diff
@@ -0,0 +1,74 @@
...
+promise agent systemd
+{
+    path        => "$(this.promise_dirname)/systemd.py";
+    interpreter => "/usr/bin/python3";
+}
+
+bundle agent test
+{
+  meta:
+    "description" -> { "CFE-3396" }
```
nickanderson, line 24: :+1:

nickanderson: I was going to say that ideally, this would also leverage dcs.cf.sub (policy to assist in writing tests in CFEngine policy DSL) from core, but then I realized this is a completely different repository and unsure if we want to bring that in.

nickanderson:
Fabio, you asked for *ideal*. I tried to type this without trying to consider what is or isn't currently possible or word conflict::

    bundle agent __main__
    {
        systemd:
          "sample"
            unit => unit( "My sample service" )
            state => unit_state( "enabled", "started" ),
            service => simple_sleep(86400),
            install => unit_install( '[ "multi-user.target" ]' );
    
        # Uninstall a unit (remove all traces)
          "sample"
            absent => "true"; # probably should error or warn if there are
                              # conflicting things like also trying to set to
                              # enabled and running
    
    }
    body unit unit( description )
    {
        description => "$(description)";
    } 
    body install unit_install(wanters)
    {
        wanted_by => { "@(wanters)" }; 
    }
    body service simple_sleep(seconds)
    {
        exec_start => "/usr/bin/sleep $(seconds)";
        type => "simple";
    }    
    body state unit_state( boot_state, service_state )
    {
       boot_state => "$(boot_state)"; # (enabled|disabled)
       unit_state => "$(service_state)"; # (started|stopped|masked)
    }    

It's so much at the top level of the promise, I kind of want to put the things in the config file together:

    bundle agent __main__
    {
        systemd:
          "sample"
            state => unit_state( "enabled", "started" ),
            config  => simple_sleep_unit( 86400 ),
            comment => "My sample service",
    
        # Uninstall a unit (remove all traces)
          "sample"
            absent => "true"; # probably should error or warn if there are
                              # conflicting things like also trying to set to
                              # enabled and running
    }
    
    body config simple_sleep_unit( seconds )
    {
       #[Unit]
        unit_description => "$(this.promise_comment)";
    
       #[Service]
        service_exec_start => "/usr/bin/sleep $(seconds)";
        service_type => "simple";
    
       #[Install]  
        install_wanted_by => { "multi-user.target" }
    } 
    
    body state unit_state( boot_state, service_state )
    {
       boot_state => "$(boot_state)"; # (enabled|disabled)
       unit_state => "$(service_state)"; # (started|stopped|masked)
    }    

However, it's my understanding that *bodies* aren't yet supported for custom promise types.

If they were supported, probably we would need to prefix the bodies so as to avoid conflict, and the pattern has been attribute name == body type which would make the top level attributes ugly.

    bundle agent __main__
    {
        systemd:
          "sample"
            systemd_unit_state => unit_state( "enabled", "started" ),
            systemd_unit_config  => simple_sleep_unit( 86400 ),
            comment => "My sample service",
    
        # Uninstall a unit (remove all traces)
          "sample"
            absent => "true"; # probably should error or warn if there are
                              # conflicting things like also trying to set to
                              # enabled and running
    }
    
    body systemd_unit_config simple_sleep_unit( seconds )
    {
       #[Unit]
        unit_description => "$(this.promise_comment)";
    
       #[Service]
        service_exec_start => "/usr/bin/sleep $(seconds)";
        service_type => "simple";
    
       #[Install]  
        install_wanted_by => { "multi-user.target" }
    } 
    
    body systemd_unit_state unit_state( boot_state, service_state )
    {
       boot_state => "$(boot_state)"; # (enabled|disabled)
       unit_state => "$(service_state)"; # (started|stopped|masked)
    }    

Maybe if/when we implement support for bodies in custom promise types, there could be something to improve the readability like:

    bundle agent __main__
    {
        systemd:
          "sample"
            state => generic( "enabled", "started" ),
            config  => simple_sleep_unit( 86400 ),
            comment => "My sample service",
    
        # Uninstall a unit (remove all traces)
          "sample"
            absent => "true"; # probably should error or warn if there are
                              # conflicting things like also trying to set to
                              # enabled and running
    }
    
    body promise_type:systemd.config simple_sleep_unit( seconds )
    {
       #[Unit]
        unit_description => "$(this.promise_comment)";
    
       #[Service]
        service_exec_start => "/usr/bin/sleep $(seconds)";
        service_type => "simple";
    
       #[Install]  
        install_wanted_by => { "multi-user.target" }
    } 
    
    body promise_type:systemd.state generic( boot_state, service_state )
    {
       boot_state => "$(boot_state)"; # (enabled|disabled)
       unit_state => "$(service_state)"; # (started|stopped|masked)
    }    

Anyway, I am unsure how helpful this is to you. Perhaps the most help is just seeing that at the top level I am trying to concisely express the desired state. No, not all the details are visible, but the most important info is.

promise_types/systemd/README.md
```diff
@@ -0,0 +1,114 @@
...
+## Attributes
+
+| Name | Type | Description| Mandatory | Default |
+| --- | --- | --- | --- | --- |
+| `daemon_reexec` | `boolean` | Run daemon-reexec before performing actions to the service | No | `false` |
+| `daemon_reload` | `boolean` | Run daemon-reload before performing actions to the service | No | `false` |
+| `enabled` | `boolean` | Start the service on boot | No | `true` |
+| `masked` | `boolean` | Mask the service, making it impossible to start | No | `false` |
+| `name` | `string` | The name of the service | Yes | Promiser |
+| `replace` | `boolean` | Replace service unit file if it already exists | No | `False` |
```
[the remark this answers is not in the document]
olehermanse: I think it's good to give the option, but default should be to fix.

promise_types/systemd/README.md
tranchitella: moved to slist

promise_types/systemd/systemd.py
tranchitella: this is a list comprehension, I cannot move the for outside it.

promise_types/systemd/systemd.py
tranchitella: it will close the file automatically, because it is not assigned to a variable

promise_types/systemd/README.md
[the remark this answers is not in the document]
tranchitella: replace = true is the default now

promise_types/systemd/systemd.py
tranchitella: Done as you suggested: we now capture stdout & stderr, and log only when actions are performed.

promise_types/systemd/README.md
tranchitella: The minimum Python version is the one set by cfbs, I don't think we need to add anything here.

tranchitella:
@olehermanse @vpodzime ready for another review round

promise_types/systemd/systemd.py
```diff
@@ -0,0 +1,394 @@
...
+        if (
+            service_status["ActiveState"] == "active"
+            and service_status["SubState"] == "running"
+        ):
+            try:
+                self._exec_command(["systemctl", "stop", model.name])
+            except subprocess.CalledProcessError as e:
+                self.log_error(f"Failed to run systemctl: {e.output or e}")
+                e.stderr and self.log_error(e.stderr.strip())
+                return (Result.NOT_KEPT, [f"{safe_promiser}_daemon_reexec_failed"])
```
olehermanse, line 148:
```suggestion
                return (Result.NOT_KEPT, [f"{safe_promiser}_daemon_stop_failed"])
```

promise_types/systemd/systemd.py
```diff
@@ -0,0 +1,394 @@
...
+            self.log_info(f"Stopped the service {model.name}")
+        # disable the service, in case it is enabled
+        if service_status["ActiveState"] == "active":
+            try:
+                self._exec_command(["systemctl", "disable", model.name])
+            except subprocess.CalledProcessError as e:
+                self.log_error(f"Failed to run systemctl: {e.output or e}")
+                e.stderr and self.log_error(e.stderr.strip())
+                return (Result.NOT_KEPT, [f"{safe_promiser}_daemon_reexec_failed"])
```
olehermanse, line 157:
```suggestion
                return (Result.NOT_KEPT, [f"{safe_promiser}_daemon_disable_failed"])
```

olehermanse (approved):
Looks very good overall. I'm sure there will be some things we want to change as we test it more, but I think we can merge it now. @tranchitella please address my comments with a follow-up PR.

promise_types/systemd/systemd.py
tranchitella: https://github.com/cfengine/modules/pull/5

promise_types/systemd/systemd.py
tranchitella: https://github.com/cfengine/modules/pull/5

PR #6, opened 2021-06-01
CFE-3680: Validation of attributes without pydantic

olehermanse:
**TODO / Follow ups:**

- [x] Remove pydantic dependency from git promise type
- [x] Remove pydantic dependency from systemd promise type: #7 
- [ ] Update cfengine.py duplicate in other repo


libraries/python/cfengine.py
```diff
@@ -39,6 +40,26 @@ def _would_log(level_set, msg_level):
     return _LOG_LEVELS[msg_level] <= _LOG_LEVELS[level_set]
 
 
+def _cfengine_type(typing):
+    if typing is str:
+        return "string"
+    if typing is int:
+        return "int"
+    if typing is list:
+        return "slist"
```
vpodzime, line 49: I guess the same should be for `tuple` just in case?

libraries/python/cfengine.py
```diff
@@ -161,13 +182,75 @@ def _add_traceback_to_response(self):
...
+        attribute["default"] = default
+        attribute["required"] = required
+        attribute["default_to_promiser"] = default_to_promiser
+        attribute["validator"] = validator
+        self._validator_attributes[name] = attribute
+
+    def _has_validation_attributes(self):
+        if self._validator_attributes:
+            return True
+        return False
```
vpodzime, line 198:
```
@property
def _has_validation_attributes(self):
    return bool(self._validator_attributes)
```

libraries/python/cfengine.py
```diff
@@ -161,13 +182,75 @@ def _add_traceback_to_response(self):
         logs.append({"level": "debug", "message": trace})
         self._response["log"] = logs
 
+    def add_attribute(self, name, typing, default=None, required=False, default_to_promiser=False, validator=None):
+        attribute = OrderedDict()
```
vpodzime, line 186: I don't understand why this is an `OrderedDict`.

libraries/python/cfengine.py
```diff
@@ -161,13 +182,75 @@ def _add_traceback_to_response(self):
...
+            expected = _cfengine_type(self._validator_attributes[name]["typing"])
+            found = _cfengine_type(type(value))
+            if found != expected:
+                raise ValidationError(f"Wrong type for attribute '{name}', requires '{expected}', not '{found}'")
+            if self._validator_attributes[name]["validator"]:
+                # Can raise ValidationError:
+                self._validator_attributes[name]["validator"](value)
+
+        # Construct as dict first:
+        attribute_dict = OrderedDict()
```
vpodzime, line 223: I don't understand why this is an `OrderedDict` either.

libraries/python/cfengine.py
```diff
@@ -39,6 +40,26 @@ def _would_log(level_set, msg_level):
...
+        return "data container"
+    if typing is bool:
+        return "true/false"
+    return "Error in promise module"
+
+
+class AttributeObject(object):
+    def __init__(self, d):
+        for key, value in d.items():
+            setattr(self, key, value)
```
vpodzime, line 60: It might be better to inherit from `dict` (or `OrderedDict`) here and only override the `__getattribute__` method. That way, `AttributeObject` could be used as a dictionary while its items could still be accessed as attributes.

libraries/python/cfengine.py
```diff
@@ -161,13 +182,75 @@ def _add_traceback_to_response(self):
...
+
+        # Copy attributes specified by user policy:
+        for key, value in attributes.items():
+            attribute_dict[key] = value
+
+        # Set defaults based on promise module validation hints:
+        for name, value in self._validator_attributes.items():
+            if value.get("default_to_promiser", False):
+                attribute_dict.setdefault(name, promiser)
+            elif value.get("default", None) != None:
```
vpodzime, line 233: `is not None`

promise_types/git/git.py
```diff
@@ -4,61 +4,39 @@
...
+        def destination_must_be_absolute(v):
+            if not os.path.isabs(v):
+                raise ValidationError(f"must be an absolute path, not '{v}'")
+
+        def depth_must_be_positive(v):
+            if not v >= 0:
+                raise ValidationError(f"must be a positive number, not '{v}'")
+
+        self.add_attribute("destination", str, default_to_promiser=True, validator=destination_must_be_absolute)
+        self.add_attribute("repository", str, required = True)
```
vpodzime, line 23: No spaces around `=` in kwargs, please.

promise_types/git/git.py
```diff
@@ -4,61 +4,39 @@
...
-            errors = [
-                ".".join(map(str, err["loc"])) + ": " + err["msg"] for err in e.errors()
-            ]
-            raise ValidationError(", ".join(errors))
+        def destination_must_be_absolute(v):
+            if not os.path.isabs(v):
+                raise ValidationError(f"must be an absolute path, not '{v}'")
+
+        def depth_must_be_positive(v):
+            if not v >= 0:
```
vpodzime, line 19: `if v < 0` might be easier to read here. Also, it's non-negative, not positive.

vpodzime (commented):
Looks good to me otherwise.

libraries/python/cfengine.py
olehermanse: Not necessary, but 2 reasons:

- Consistency since all dicts are ordered in later versions of python
- Nice to preserve order when iterating over or printing attributes from policy. I.e. the attributes the user typed explicitly come first, in the order they typed it, while defaults come after.

libraries/python/cfengine.py
olehermanse: Same as in the other case.

libraries/python/cfengine.py
olehermanse: Okay, but this comes from the json parser, which creates lists.

libraries/python/cfengine.py
olehermanse: Done.

libraries/python/cfengine.py
olehermanse: I will refer to [PEP 20](https://www.python.org/dev/peps/pep-0020/) here:

> There should be one-- and preferably only one --obvious way to do it.

libraries/python/cfengine.py
olehermanse: Done.

promise_types/git/git.py
olehermanse: Done.

promise_types/git/git.py
olehermanse: Done.

libraries/python/cfengine.py
olehermanse: (Fixed)

olehermanse:
@vpodzime see force push: https://github.com/cfengine/modules/compare/98a53ea63f931dd2dd165600512e0dd7a7441bfd..eedd994651b0a48a8ea2893af7e39dabb86ed90a

vpodzime (approved):
ACK

PR #7, opened 2021-06-02
CFE-3680: Removed pydantic dependency from systemd promise type

olehermanse:


promise_types/systemd/example.cf
```diff
@@ -14,7 +14,7 @@ bundle agent main
       enabled => "true",
       state => "restarted",
       unit_description => "my sample service",
-      service_exec_start => "/usr/bin/sleep 86400",
+      service_exec_start => {"/usr/bin/sleep 86400"},
```
vpodzime, line 17: Looks like this should be part of the first commit.

promise_types/systemd/test.cf
```diff
@@ -31,7 +31,7 @@ bundle agent test
       enabled => "true",
       state => "restarted",
       unit_description => "my sample service",
-      service_exec_start => "/usr/bin/sleep 86400",
+      service_exec_start => {"/usr/bin/sleep 86400"},
```
vpodzime, line 34: Should also be part of the first commit.

promise_types/systemd/systemd.py
```diff
@@ -24,86 +19,73 @@ class SystemdPromiseTypeStates(Enum):
...
+
+        self.add_attribute("daemon_reexec", bool, default=False)
+        self.add_attribute("daemon_reload", bool, default=False)
+        self.add_attribute("enabled", bool, default=True)
+        self.add_attribute("masked", bool, default=False)
+        self.add_attribute("name", str, required=True, default_to_promiser=True)
+        self.add_attribute("replace", bool, default=True)
+        self.add_attribute("state", str, required=True, validator=state_must_be_valid)
+        self.add_attribute("unit_description", str)
+        self.add_attribute("unit_requires", list, default=[])
```
vpodzime, line 47: Beware using `[]` as the default value for a keyword argument (the list will only be created once). But I think it's ok here.

vpodzime (commented):
Looks good to me otherwise.

promise_types/systemd/example.cf
olehermanse: Fair enough, but IIRC when I added the example the pydantic validation accepted it.

promise_types/systemd/systemd.py
olehermanse: I added a commit to always copy the default: https://github.com/cfengine/modules/pull/7/commits/9b25e362bc21eac1b8115e2150c4977f406be0b8

promise_types/systemd/test.cf
olehermanse: Fixed.

promise_types/systemd/example.cf
olehermanse: Fixed.

libraries/python/cfengine.py
```diff
@@ -135,6 +136,30 @@ def start(self, in_file=None, out_file=None):
...
+            # Only known conversion needed: "true"/"false" -> True/False
+            if self._validator_attributes[name]["typing"] is bool:
+                if value == "true":
+                    replacements[name] = True
+                elif value == "false":
+                    replacements[name] = False
+
+        # Don't edit dict while iterating over it, after instead:
+        for k, v in replacements.items():
+            attributes[k] = v
```
tranchitella, line 158:
```suggestion
        attributes.update(replacements)
```

tranchitella (approved):

PR #8, opened 2021-06-03
cfengine.py: Bring in changes from core

olehermanse:
See:
https://github.com/cfengine/core/pull/4697
https://github.com/cfengine/core/pull/4700

### DIFF
--- a/libraries/python/cfengine.py
+++ b/libraries/python/cfengine.py
@@ -146,6 +146,9 @@ class PromiseModule:
             if type(value) is not str:
                 # If something is not string, assume it is correct type
                 continue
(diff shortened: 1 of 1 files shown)


tranchitella (approved):

vpodzime:
This ping-pong is really weird. We should decide which place we make changes at and which place will just pull them.

olehermanse:
> This ping-pong is really weird. We should decide which place we make changes at and which place will just pull them.

@vpodzime Sorry, I wanted to make all the changes in modules repo and bring them to core. But it turns out that in core we have pesky tests that find bugs in my code, so here we are 🤷

PR #9, opened 2021-06-04
CFE-3655: ansible promise type for running playbooks

tranchitella:


tranchitella:
@olehermanse ping

promise_types/ansible/README.md
```diff
@@ -0,0 +1,79 @@
...
+* Ansible >= 2.8.0
+
+## Attributes
+
+| Name | Type | Description| Mandatory | Default |
+| --- | --- | --- | --- | --- |
+| `playbook` | `string` | Absolute path of the Ansible playbook | No | Promiser |
+| `inventory` | `string` | Absolute path of the inventory file | No | - |
+| `limit` | `slist` | List of host names to target | No | `{"localhost"}` |
+| `tags` | `slist` | List of tags to play | no | `{}` |
```
olehermanse, line 20:
```suggestion
| `tags` | `slist` | List of tags to play | No | `{}` |
```

promise_types/ansible/README.md
```diff
@@ -0,0 +1,79 @@
...
+| `become` | `boolean` | Set the `become` option | - | `False` |
+| `become_method` | `string` | Set the `become_method` option | - | `"sudo"` |
+| `become_user` | `string` | Set the `become_user` option | - | `root` |
+| `connection` | `string` | Set the `connection` option; possible values: `local`, `ssh` | - | `local` |
+| `forks` | `int` | Set the `forks` option | - | `1` |
+| `private_key_file` | `string` | Absolute path of the SSH private key to use | - | - |
+| `remote_user` | `string` | Set the `remote_user` option | - | `root` |
```
olehermanse, line 27: Why do these have ` - ` for Mandatory column while the ones above have `No` ?

promise_types/ansible/README.md
```diff
@@ -0,0 +1,79 @@
...
+      limit            => {"host1", "host2", "host3", "host4"},
+      private_key_file => "/path/to/your/private/key/id_rsa";
+}
+```
+
+## Authors
+
+This software was created by the team at [Northern.tech AS](https://northern.tech), with many contributions from the community. Thanks everyone!
+
+[CFengine](https://cfengine.com) is sponsored by [Northern.tech AS](https://northern.tech)
```
olehermanse, line 69:
```suggestion
[CFEngine](https://cfengine.com) is sponsored by [Northern.tech AS](https://northern.tech)
```

promise_types/ansible/ansible_promise.py
```diff
@@ -0,0 +1,163 @@
...
+                    "Summary of the tasks for '" + host + "' is: " + summary
+                )
+
+    ANSIBLE_AVAILABLE = True
+except ImportError:
+    raise
+    ANSIBLE_AVAILABLE = False
+
+
+COMPAT_OPTIONS = (
```
olehermanse, line 59: Maybe a comment to explain what COMPAT_OPTIONS means? Compatibility with what?

promise_types/ansible/example.cf
```diff
@@ -0,0 +1,15 @@
...
+  path => "$(sys.workdir)/modules/promises/ansible_promise.py";
+  interpreter => "/usr/bin/python3";
+}
+
+bundle agent main
+{
+  ansible:
+    "my_playbook"
+      playbook  => "/tmp/playbook.yaml",
+      inventory => "/tmp/inventory.yaml",
```
olehermanse, line 13: Maybe you could add a small dummy playbook to make it complete and the example runnable?

promise_types/ansible/ansible_promise.py
```diff
@@ -0,0 +1,163 @@
...
+
+class AnsiblePromiseTypeModule(PromiseModule):
+    def __init__(self, **kwargs):
+        super(AnsiblePromiseTypeModule, self).__init__(
+            "ansible_promise_module", "0.1.0", **kwargs
+        )
+
+        def must_be_absolute(v):
+            if not os.path.isabs(v):
+                raise ValidationError(f"must be an absolute path, not '{v}'")
```
olehermanse, line 77:
```suggestion
                raise ValidationError(f"Must be an absolute path, not '{v}'")
```

olehermanse (approved):
Looks really good, Fabio! I haven't tested it, but here are some smaller suggestions.

olehermanse:
@tranchitella your change to cfengine.py was already merged here: https://github.com/cfengine/modules/pull/10

You can rebase and drop that commit :)

promise_types/ansible/README.md
tranchitella: It's a mistake, I'm fixing it in the other READMEs as well.

promise_types/ansible/ansible_promise.py
tranchitella: sorry, it was a left-over

tranchitella:
@olehermanse I addressed your comments, rebased and fixed the other README files.
Can you please review it again and merge, in case it is okay? I don't have merge rights in this repo.
Thanks.

promise_types/ansible/ansible_promise.py
```diff
@@ -0,0 +1,164 @@
...
+
+try:
+    from ansible import context
+    from ansible.cli import CLI
+    from ansible.executor.playbook_executor import PlaybookExecutor
+    from ansible.inventory.manager import InventoryManager
+    from ansible.module_utils.common.collections import ImmutableDict
+    from ansible.parsing.dataloader import DataLoader
+    from ansible.plugins.callback import CallbackBase
+    from ansible.vars.manager import VariableManager
```
vpodzime, line 15: Please only do the imports in the `try-except` block. The class definition shouldn't be inside it.

promise_types/ansible/ansible_promise.py
```diff
@@ -0,0 +1,164 @@
...
+        safe_promiser = promiser.replace(",", "_")
+        return (safe_promiser, attributes)
+
+    def validate_promise(self, promiser: str, attributes: Dict):
+        if not ANSIBLE_AVAILABLE:
+            raise ValidationError("Ansible Python module not available")
+
+    def evaluate_promise(
+        self, safe_promiser: str, attributes: Dict
+    ) -> Tuple[str, List[str]]:
```
vpodzime, line 104: This formatting breaks the flow of the code. I think this would be better:
```
def evaluate_promise(self, safe_promise: str, attributes: Dict)
                                   -> Tuple[str, List[str]]:
```
or even with everything on one line. Nobody's going to print this code on a toilet paper.

promise_types/ansible/ansible_promise.py
```diff
@@ -0,0 +1,164 @@
...
+        )
+        pbex = PlaybookExecutor(
+            playbooks=[attributes["playbook"]],
+            inventory=inventory,
+            variable_manager=variable_manager,
+            loader=loader,
+            passwords={},
+        )
+        callback = CallbackModule(promise=self)
+        pbex._tqm._stdout_callback = callback
```
vpodzime, line 151: Is there a way to avoid accessing private attributes?

vpodzime (changes requested):
Please split the changes in the *git* module into a separate commit.

promise_types/ansible/ansible_promise.py
tranchitella: The class depends on CallbackBase, which is imported conditionally. This is the reason why the class definition is in the try block. This is a common pattern in Python, I believe.

promise_types/ansible/ansible_promise.py
tranchitella: The code is automatically formatted using black to avoid formatting bikeshedding.

promise_types/ansible/ansible_promise.py
tranchitella: > Please split the changes in the _git_ module into a separate commit.

done

tranchitella: > Is there a way to avoid accessing private attributes?

Yes, but it would require bloating the Python code with all the initialization of the task queue manager which the PlaybookExecutor embeds. In any case, _stdout_callback is private, thus we would need to override the TaskQueueManager, adding additional complexity. I don't think it's a good idea, but if you disagree I can give it a try.

promise_types/ansible/ansible_promise.py
vpodzime: The class can be defined inside an `if ANSIBLE_AVAILABLE`.

promise_types/ansible/ansible_promise.py
vpodzime: Not worth the struggle, AFAICT. We just need to hope the names stay the same.

promise_types/ansible/ansible_promise.py
tranchitella: ok done

promise_types/ansible/ansible_promise.py
tranchitella: I'm use black to format my Python code, if the policy in CFEngine is not relying on autoformatter let me know and I'll manually format it according to your policy.

olehermanse: @tranchitella black is better ;) Ignore him, your formatting is perfect.

olehermanse: @vpodzime The problem is that you end up arguing something extremely subjective. I don't print my code on toilet paper and I still think Fabios snippet here is more readable than your suggestion.

vpodzime: OK. If **you guys** prefer such formatting, that's the way to go. But not just because some formatter puts it that way. Formatting preferences differ, for me "everything after `def something` should be indented more than the `def`" is quite a key for readability and flow of the code. But YMMV, of course.

PR #10, opened 2021-06-04
cfengine.py: int attributes and critical error message for bug in promise module

olehermanse:


olehermanse:
Since these were already reviewed in core, I will merge.

PR #12, opened 2021-07-27
ENT-7423: Added iptables custom promise module concept DRAFT

KotzaBoss:
Ticket: ENT-7423

### DIFF
--- a/promise_types/iptables/README.md
+++ b/promise_types/iptables/README.md
@@ -1,0 +1,198 @@
+# Iptables Promise Module
+
+## Synopsis
+- *Name*: `iptables`
+- *Version*: `0.1.1`
+- *Description*: Manage network packet filter rules
+
+## Requirements
+- [`iptables`](https://manpages.ubuntu.com/manpages/precise/en/man8/iptables.8.html) (command line tool)
+
+## Attributes
+
+| Name        | Type  | Description                                                                                     | Mandatory | Default    |
+|:------------|-------|-------------------------------------------------------------------------------------------------|-----------|------------|
+| `command`   | `str` | Which command should be run. Depending on the choice different attributes are accepted          | Yes       | -          |
+| `table`     | `str` | Table to operate on                                                                             | No        | `"filter"` |
+| `chain`     | `str` | Chain to operate on. `"ALL"` can be used in commands like flush, to flush all chains of a table | Yes       | -          |
+| `rulenum`   | `int` | Index to put the new rule with `1` meaning it will be inserted at the top                       | No        | `1`        |
+| `protocol`  | `str` | Protocol to be used                                                                             | No        | -          |
+| `dest_port` | `int` | Tcp/udp network port                                                                            | No        | -          |
+| `source_ip` | `str` | Source IP of incoming traffic                                                                   | No        | -          |
+| `target`    | `str` | Target for packet if the rule matches                                                           | No        | -          |
+
+## Command Validation
+
+| *Command\Valid Attribute* | **table** | **chain** | **rulenum** | **protocol** | **dest_port** | **source_ip** | **target** |
+|---------------------------|-----------|-----------|-------------|--------------|---------------|---------------|------------|
+| **append**                | Yes       | Yes       | No          | Yes          | Yes           | Yes           | Yes        |
+| **insert**                | Yes       | Yes       | Yes         | Yes          | Yes           | Yes           | Yes        |
+| **delete**                | Yes       | Yes       | No          | Yes          | Yes           | Yes           | Yes        |
+| **flush**                 | Yes       | Yes       | No          | No           | No            | No            | No         |
+| **policy**                | Yes       | Yes       | No          | No           | No            | No            | Yes        |
+
+Its important to note that some attribute have their own validation requirements. For example `dest_port` should expect `protocol` to be present.
+
+## Examples
+
+### Append rule to accept traffic from some IP
+
+```cfengine3
+bundle agent main
+{
+  iptables:
+      "accept_cfengine_com"
+        command => "append",
+        chain => "INPUT",
+        source_ip => "34.107.174.45",
+        target => "ACCEPT";
+}
+```
+
+### Append a rule to drop traffic from telnet port
+
+```cfengine3
+bundle agent main
+{
+  iptables:
+      "drop_telnet"
+        command => "append",
+        chain => "INPUT",
+        protocol => "tcp",
+        dest_port => 23,
+        target => "DROP";
+}
+```
+
+### Delete a rule that accepts traffic from some IP
+
+```cfengine3
+bundle agent main
+{
+  iptables:
+      "delete_accept_cfengine_com"
+        command => "delete",
+        chain => "INPUT",
+        source_ip => "34.107.174.45",
+        target => "ACCEPT";
+}
+```
+
+###  Insert a rule to accept ssh port as the first rule
+
+```cfengine3
+bundle agent main
+{
+  iptables:
+      "accept_ssh"
+        command => "insert",
+        chain => "INPUT",
+        rulenum => 1,
+        protocol => "tcp",
+        dest_port => 22,
+        target => "ACCEPT";
+}
+```
+
+### Set an aggressive policy for incoming traffic
+
+```cfengine3
+bundle agent main
+{
+  iptables:
+      "aggressive_policy"
+        command => "policy",
+        chain => "INPUT",
+        target => "DROP";
+}
+```
+
+### Flush a specific filter chain
+
+``` cfengine3
+bundle agent main
+{
+  iptables:
+      "flush_INPUT"
+        command => "flush",
+        chain => "INPUT";
+}
+```
+
+### Flush all filter chains
+
+```cfengine3
+bundle agent main
+{
+  iptables:
+      "flush_all"
+        command => "flush",
+        chain => "ALL";
+}
+```
+
+### Accept traffic from CFEngine port, drop everything else
+
+``` cfengine3
+bundle agent main
+{
+  iptables:
+      "flush_all"
+        command => "flush",
+        chain => "ALL";
+
+      "aggressive_policy"
+        command => "policy",
+        chain => "INPUT",
+        target => "DROP";
+
+      "accept_cfengine"
+        command => "append",
+        chain => "INPUT",
+        protocol => "tcp",
+        dest_port => 5308,
+        target => "ACCEPT";
+}
+```
+## Notes
+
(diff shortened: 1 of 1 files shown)


vpodzime:
Like I said before -- I think we should just mirror the [_iptables Ansible module_](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/iptables_module.html). It's not perfect, but will/might be familiar with it already. No reason to have _command_ instead of _action_ etc.

KotzaBoss:
@vpodzime 
> Like I said before

I havent forgotten what you mentioned, im just unsure about how far "mirroring" goes before its against their GPL3 license. Im currently doing more research on licenses because its something im not very well educated in.

For the current attributes i have listed, there is primarily what you mentioned that deviates besides a trivial rename from `jump` to `target` (afaict).

Something that they have that i didnt follow is the `state` attribute. That made me double take a bit because in my minds eye the `command` i have atm more or less merges ansibles `action`/`state`. At the same time having that "command" also centralizes validation as for example in ansible `flush` is "mutualy exclusive" with `policy`. 

That is my reasoning for avoiding that particular attribute setup from ansible. If you consider "complete" mirroring of ansible a preferable choice i can easily convert my draft to that (though as i said im still unsure about licenses).

vpodzime:
> I havent forgotten what you mentioned, im just unsure about how far "mirroring" goes before its against their GPL3 license. Im currently doing more research on licenses because its something im not very well educated in.

@olehermanse perhaps we should have a *Licensing* section somewhere in the docs. Also, I think we should extend the *LICENSE* file in this repo to mention that individual modules can use various licenses.?

promise_types/iptables/README.md
```diff
@@ -0,0 +1,153 @@
...
+                  iptables -t filter -F";
+ 		            
+    "flush_INPUT"
+      command => "flush",
+      chain => "INPUT",
+      comment => "Flush INPUT chain for filter:
+                  iptables -t filter -F INPUT"
+```
+
+It should be noted that the commands shown in the comments will not be blindly executed. Appending, inserting and deleting will first run `iptables --check` in the background to ensure that the promise is not kept before executing the commands. Policy and flush are more lenient.
```
nickanderson, line 128: How does `--check` work with `insert`?

```
root:~# iptables --check filter -I 1 INPUT -p tcp --dport 22 -j ACCEPT
iptables v1.4.21: Cannot use -I with -C

```

KotzaBoss: Very clunky expression on my part. I fixed it [here](https://github.com/KotzaBoss/modules/blob/iptables-custom-promise/promise_types/iptables/README.md#notes)

nickanderson: Can check validate that a rule is in a certain position (rulenum) ?

KotzaBoss: Check doesnt accept any rule num unfortunately. To find the rulenum you can use the flag `--line-numbers` with `--list` like so
```shell
> iptables -L --line-numbers
Chain INPUT (policy ACCEPT)
num  target     prot opt source               destination         
1    ACCEPT     all  --  anywhere             anywhere            
2    ACCEPT     all  --  anywhere             anywhere            
3    ACCEPT     all  --  anywhere             anywhere            

Chain FORWARD (policy ACCEPT)
num  target     prot opt source               destination         

Chain OUTPUT (policy ACCEPT)
num  target     prot opt source               destination 
```
But this is obviously a more manual way of finding the rulenum because you would have to parse the output (which isnt that hard just abit clunky). It would be nice to have a command like check that has as return code the position of the rule but it doesnt exist afaik.

promise_types/iptables/README.md
```diff
@@ -0,0 +1,179 @@
...
+      3. [Inserting](#inserting)
+      4. [Policy](#policy)
+      5. [Flush](#flush)
+   4. [Notes](#notes)
+2. [Appendix](#appendix)
+   1. [Man iptables](#man-iptables)
+
+# Iptables Custom Promise Draft
+
+The expected structure of `iptables` [commands](#man-iptables) is fairly well defined. This can work to our advantage when making the attribute validation. This draft presents a preliminary concept of how adding/deleting rules, flushing chains and setting policies could look like.
```
olehermanse, line 19: Please copy the README from here for consistency:

https://github.com/cfengine/modules/blob/master/promise_types/git/README.md

promise_types/iptables/README.md
```diff
@@ -0,0 +1,179 @@
...
+
+## Command Validation
+
+| *Command\Valid Attribute* | table      | rulenum    | chain      | protocol   | dest_port  | source_ip  | target     |
+| ------------------------- | ---------- | ---------- | ---------- | ---------- | ---------- | ---------- | ---------- |
+| **append**                | ![][green] | ![][red]   | ![][green] | ![][green] | ![][green] | ![][green] | ![][green] |
+| **insert**                | ![][green] | ![][green] | ![][green] | ![][green] | ![][green] | ![][green] | ![][green] |
+| **delete**                | ![][green] | ![][red]   | ![][green] | ![][green] | ![][green] | ![][green] | ![][green] |
+| **flush**                 | ![][green] | ![][red]   | ![][green] | ![][red]   | ![][red]   | ![][red]   | ![][red]   |
+| **policy**                | ![][green] | ![][red]   | ![][green] | ![][red]   | ![][red]   | ![][red]   | ![][green] |
```
olehermanse, line 42: Red-green colorblindness is quite common. Just write _yes_ / _no_ instead.

promise_types/iptables/README.md
```diff
@@ -0,0 +1,179 @@
...
+| `rulenum`   | `int` | Integer value greater or equal to `1`                        | Index to put the new rule. 1 means it will be inserted at the top. | `1`        |
+| `chain`     | `str` | `"INPUT"`, `"OUTPUT"`, `"FORWARD"`, `"PREROUTING"`, `"POSTROUTING"`, `"ALL"` | Chain to operate on. `"ALL"` can be used in commands like flush to flush all chains of a table. | -          |
+| `protocol`  | `str` | `"tcp"`, `"udp"`                                             | Protocol to be used.                                         | -          |
+| `dest_port` | `int` | `0 - 65535`                                                  | Tcp/udp network port                                         | -          |
+| `source_ip` | `str` | Network IP address                                           | Source IP of incoming traffic.                               | -          |
+| `target`    | `str` | `"ACCEPT"`, `"DROP"`, `"REJECT"`                             | Target for packet if the rule matches.                       | -          |
```
olehermanse, line 32: Please use the same columns as for other modules - Name, type, description, mandatory, default.

https://github.com/cfengine/modules/blob/master/promise_types/git/README.md

promise_types/iptables/README.md
```diff
@@ -0,0 +1,179 @@
...
+| **insert**                | ![][green] | ![][green] | ![][green] | ![][green] | ![][green] | ![][green] | ![][green] |
+| **delete**                | ![][green] | ![][red]   | ![][green] | ![][green] | ![][green] | ![][green] | ![][green] |
+| **flush**                 | ![][green] | ![][red]   | ![][green] | ![][red]   | ![][red]   | ![][red]   | ![][red]   |
+| **policy**                | ![][green] | ![][red]   | ![][green] | ![][red]   | ![][red]   | ![][red]   | ![][green] |
+
+Its important to note that some attribute have their own validation requirements. For example `dest_port` should expect `protocol` to be present. To maintain focus in this draft that kind of validation is left out.
+
+## Examples
+
+### Appending
```
olehermanse, line 48: I think the title of each example should mention the high level use case.

For example: "Appending a rule to drop telnet traffic"

promise_types/iptables/README.md
```diff
@@ -0,0 +1,179 @@
...
+                  iptables -t filter -A INPUT -p tcp --dport 5308 -j ACCEPT";
+
+    "drop_some_ip"
+      command => "append",
+      # table omitted
+      chain => "INPUT",
+      source_ip => "1.2.3.4",
+      target => "DROP",
+      comment => "Drop traffic from ip 1.2.3.4:
+                  iptables -t filter -A INPUT -s 1.2.3.4 -j DROP";
```
olehermanse, line 69: Prefer shorter examples. The first example should be as short as possible.

promise_types/iptables/README.md
```diff
@@ -0,0 +1,179 @@
...
+
+```cfengine3
+iptables:
+    "delete_drop_some_id"
+      command => "delete",
+      chain => "INPUT",
+      source_ip => "1.2.3.4",
+      target => "DROP",
+      comment => "Delete the rule for ip 1.2.3.4:
+                  iptables -t filter -D INPUT -s 1.2.3.4 -j DROP";
```
olehermanse, line 82: Add the 3 lines to all your examples, so I can run them by just copy pasting:

```cf3
bundle agent main
{

}
```

promise_types/iptables/README.md
```diff
@@ -0,0 +1,179 @@
...
+
+```cfengine3
+iptables:
+    "aggressive_policy"
+      command => "policy",
+      chain => "INPUT",
+      target => "DROP",
+      comment => "Drop all traffic:
+                  iptables -t filter -P INPUT DROP";
+```
```
olehermanse, line 110: After this example, it would make sense to have another one which achieves something similar; Accept port 5308 (CFEngine), drop everything else.

promise_types/iptables/README.md
```diff
@@ -0,0 +1,179 @@
...
+
+-P, --policy chain target
+```
+```
+rule-specification = [matches...] [target]
+
+match = -m matchname [per-match-options]
+
+target = -j targetname [per-target-options]
+```
```
olehermanse, line 179: Not necessary to have here (you can add a link if you want to: https://linux.die.net/man/8/iptables

olehermanse:
> I havent forgotten what you mentioned, im just unsure about how far "mirroring" goes before its against their GPL3 license. Im currently doing more research on licenses because its something im not very well educated in.

@KotzaBoss Copying the verbs action / command etc. is okay. (Of course you cannot copy their code).

promise_types/iptables/README.md
KotzaBoss: this will require the flush action too so perhaps move that at the end after examples of flush?

olehermanse: Yea, makes sense.

vpodzime:
>  (Of course you cannot copy their code).

Why not? We can license the module as GPLv3, right?

olehermanse:
> Why not? We can license the module as GPLv3, right?

Yea, but that's not the same as just copying parts of the code.

We can consider including some modules from other projects, while respecting their license, but that's not what we are doing here. George is implementing an iptables module "from scratch" without dependencies, with an MIT license, and thus should not copy code from another module.

vpodzime:
> with an MIT license, and thus should not copy code from another module

I was trying to say that if we license our new module "from scratch" without dependencies as GPLv3, we can actually copy code from the ansible module instead of re-inventing the wheel here.

PR #13, opened 2021-08-09
Fixed environment variable logic and added ssh_executable attribute

KotzaBoss:
Running the following promise with the original code will lead to errors.

```cfengine3
bundle agent main
{
  git:
    "cxxvector"
      repository => "git@github.com:KotzaBoss/CXXVector.git",
      destination => "/tmp/CXXVector",
      ssh_options => "-i /home/__alias__/.ssh/id_rsa",
      version => "master";
}
```

By putting a simple logging after setting up the environment:

```python
def _git_envvars(self, model: object):
    ...
    self.log_critical(env)
    return env
```

we see the `GIT_SSH_COMMAND` environment variable is set to the problematic value:

```
CRITICAL: { ..., 'GIT_SSH_COMMAND': 'git -i /home/__alias__/.ssh/id_rsa'}
```

The change introduces a new attribute `ssh_executable` which defaults to `'ssh'`
solving this error and allowing valid customisation of the ssh command.

Ticket: None
Changelog: None

### DIFF
--- a/promise_types/git/README.md
+++ b/promise_types/git/README.md
@@ -24,6 +24,7 @@
 | `recursive` | `boolean` | If `true`, use the `--recursive` git option | No | `yes` |
 | `reference` | `string` | If set, use the `--reference` git option with the given value | No | - |
 | `remote` | `string` | Name of the git remote | No | `origin` |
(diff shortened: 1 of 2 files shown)


tranchitella (approved):

olehermanse (approved):
Great, thanks :)

PR #17, opened 2021-08-16
CFE-3728: Refactored systemd custom promise to be Python 3.5 compatible

KotzaBoss:
Testing was done by running example.cf with python3.5 interpreter

Ticket: CFE-3728


tranchitella (approved):
Thanks a lot!

PR #18, opened 2021-08-24
ENT-7423: Added python implementation for iptables custom promise

KotzaBoss:
Ticket: ENT-7423

### DIFF
--- a/promise_types/iptables/iptables.py
+++ b/promise_types/iptables/iptables.py
@@ -1,0 +1,292 @@
+from cfengine import PromiseModule, ValidationError, Result, AttributeObject
+from typing import Callable, List, Dict, Tuple
+from collections import namedtuple
+from copy import deepcopy
+import subprocess
+
+Command = namedtuple("Command", "flag denied_attrs")
+Parameter = namedtuple("Parameter", "attr_name flag")
+
+
+class Model:
+    def __init__(
+        self,
+        attr_obj: AttributeObject,
+        *,
+        commands: Dict[str, Command],
+        parameters: Tuple[Parameter]
+    ):
+        for k, v in attr_obj.__dict__.items():
+            if v is not None:
+                setattr(self, k, str(v))
+        self._commands = commands
+        self._parameters = parameters
+
+    @property
+    def rule_spec(self) -> List[str]:
+        rule_spec = []
+        for param in self._parameters:
+            value = getattr(self, param.attr_name, None)
+            if value:
+                rule_spec.extend([param.flag, value])
+        return rule_spec
+
+    @property
+    def args(self) -> List[str]:
+        args = [self.executable]
+        args.extend(["-t", self.table])
+        args.append(self._commands[self.command].flag)
+
+        if self.command != "flush" or self.chain != "ALL":
+            args.append(self.chain)
+
+        if self.command == "policy":
+            args.append(self.target)
+        else:
+            args.extend(self.rule_spec)
+
+        return args
+
+    @property
+    def needs_check(self):
+        return self.command in {"append", "insert"}
+
+    @property
+    def log_str(self) -> str:
+        if self.command in {"append", "insert", "delete"}:
+            return "{cmd}-ing rule '{spec}' at chain '{chain}' of table '{table}' {rulenum}".format(
+                cmd=self.command.capitalize(),
+                spec=" ".join(self.rule_spec),
+                chain=self.chain,
+                table=self.table,
+                rulenum="at position '{}'".format(self.rulenum)
+                if getattr(self, "rulenum", None)
+                else "",
+            )
+
+        elif self.command == "policy":
+            return "Setting policy of chain '{chain}' in table '{table}' to '{target}'".format(
+                chain=self.chain, table=self.table, target=self.target
+            )
+
+        elif self.command == "flush":
+            return "Flushing '{}' chain(s)".format(self.chain)
+
+    @staticmethod
+    def make_check_command_model(model: "Model") -> "Model":
+        check_model = deepcopy(model)
+        check_model.command = "check"
+        return check_model
+
+    def __repr__(self):
+        return "Model({})".format(
+            ", ".join("{}={!r}".format(k, v) for k, v in self.__dict__.items())
+        )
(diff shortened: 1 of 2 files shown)


promise_types/iptables/iptables.py
```diff
@@ -0,0 +1,279 @@
...
+    "REJECT",
+    "DROP",
+}
+
+
+class Model:
+    def __init__(self, attr_obj: AttributeObject):
+        for k, v in attr_obj.__dict__.items():
+            if v is not None:
+                setattr(self, k, str(v))
```
vpodzime, line 58: Looks like `Model` should be inheriting from `AttributeObject`.

promise_types/iptables/iptables.py
```diff
@@ -0,0 +1,279 @@
...
+    Parameter("protocol", "-p"),  # Some work on differentiating "sub" options may be necessary (?) (example --dport is a match option for -p tcp)
+    Parameter("destination_port", "--dport"),
+    Parameter("target", "-j"),
+)
+
+TARGETS = {
+    "ACCEPT",
+    "REJECT",
+    "DROP",
+}
```
vpodzime, line 51: I believe these variables should be inside the respective classes not global variables.

promise_types/iptables/iptables.py
```diff
@@ -0,0 +1,279 @@
...
+    def __init__(self, attr_obj: AttributeObject):
+        for k, v in attr_obj.__dict__.items():
+            if v is not None:
+                setattr(self, k, str(v))
+
+    @property
+    def rule_spec(self) -> List[str]:
+        rule_spec = []
+        for param in PARAMETERS:
+            value = getattr(self, param.attr_name, None)
```
vpodzime, line 64: `getattr()` returns `None` by default.

promise_types/iptables/iptables.py
```diff
@@ -0,0 +1,279 @@
...
+                cmd=self.command.capitalize(),
+                spec=" ".join(self.rule_spec),
+                chain=self.chain,
+                table=self.table,
+                rulenum="at position '{}'".format(self.rulenum)
+                if getattr(self, "rulenum", None)
+                else "",
+            )
+
+        if self.command == "policy":
```
vpodzime, line 102: `elif`

promise_types/iptables/test.cf
```diff
@@ -0,0 +1,143 @@
...
+        command => "policy",
+        chain => "INPUT",
+        target => "DROP";
+
+      "accept_cfengine"
+        command => "append",
+        chain => "INPUT",
+        protocol => "tcp",
+        destination_port => "5308",
+        target => "ACCEPT";
```
vpodzime, line 139: This would be making all the changes over and over, right?

vpodzime (commented):
Some comments, but the biggest concern I have is that this is not idempotent -- it would make all the changes over and over instead of maintaining some state.

promise_types/iptables/test.cf
KotzaBoss: `insert`, `append`, `delete` do a check in the background before running the command. `flush` and `policy` dont atm.
I did want to listen to feedback but my reasoning was that setting a policy and flushing could be ran regardless of what the state of the chain is. Whereas inserting would obviously be adding the same rule over and over (hence the check first).

If checking the state before flushing or setting policy in this case is important then i can simply do that check, which however may be clunky as i may need to parse iptables input.
```shell
# After running all tests
> iptables -t filter -S INPUT
-P INPUT DROP
-A INPUT -p tcp -m tcp --dport 5308 -j ACCEPT
```
So _before_ running `flush` or `policy` i can simply parse this output and check:
- If the first line is `-P specified_chain promised_policy` for `policy`
- There is only one line in this output, meaning the chain is already `flush`

Do you think this moves us closer to how cfengine should work?

promise_types/iptables/iptables.py
KotzaBoss: I looked at the docs and it actually raises an exception if its not explicit
>  If the named attribute does not exist, default is returned if provided, otherwise AttributeError is raised.

https://docs.python.org/3.5/library/functions.html?highlight=getattr#getattr

promise_types/iptables/test.cf
olehermanse: We need things to be idempotent, or at the very least if some things are not, you should demonstrate how to use them in policy in an idempotent way. (Either the module or the policy needs to be able to check whether a change has to be made).

olehermanse: In this case, it makes some sense that the 3 different promises are not idempotent combined. (Flush, drop, accept).

However, we must find a way to achieve this use case in an idempotent way. It must be possible to write a fairly simple policy files which ensures iptables accepts CFEngine and drops everything else. And have that policy emit INFO messages only once, when it makes changes, and not in subsequent passes and agent runs.

promise_types/iptables/iptables.py
KotzaBoss: @olehermanse What you say indeed makes sense however the inheritance change i suggested wasnt about the _Iptables_ class at all. Perhaps i didnt make my example clear.
 
The change has to do exclusively with the `AttributeObject` and the `Model` classes (or any potential subclass). Of course it makes sense to have the iptables class be _composed_ of a attribute object instead of inheriting. In fact your example is smth i will definitely have in mind when thinking about composition.

What i suggested was that the logic of validation that i take advantage of through `create_attribute_object` could perhaps be moved to the `AttributeObject` itself. That way subclasses (in this case the `Model`) can take the "default" validation and extend it. In the case of the `Model`, it is not interested in any attributes that are `None`. Once again the iptables class has no idea of any of this (funnily enough it is not even composed of the model, i use it as a local variable in the `evaluate_promise` function.

promise_types/iptables/test.cf
vpodzime: Another problem here is that if you do flush and then add rules, there's a period of time in which there are no firewall rules. So a potential attack could get through it in that window of time.

olehermanse:
Closing this as stale, but we might get back to it later.

PR #22, opened 2021-09-30
Fixed value error bug in groups_experimental promise module

larsewi:
Fixed value error bug in promise module, formatted promise module with Black, and cleaned up groups_experimental policy example.


vpodzime (commented):
Please elaborate a bit more what's going on in the first commit message. Also, now you are back at this module, can you get rid of all the `f""` strings to lower the minimum required version of Python?

vpodzime (approved):
Looks good to me. Please create a ticket for changing the module to actually work with lists/slists and objects/data attributes instead of the string. (And adapting the test.)

olehermanse (approved):

PR #30, opened 2021-10-28
'http' promise type

vpodzime:


olehermanse (changes requested):
Looks promising so far. I think the lack of locking is okay for now; we can use `if` attribute etc. in policy to get around it.

Please add a policy example, README.md and enable.cf files similar to the other promise types.

olehermanse (approved):

PR #36, opened 2021-11-01
Improve README of the 'http' promise type module

vpodzime:
Ticket: ENT-7624


olehermanse (changes requested):
Please remove the README.md that I added, and use the same format as the other promise types to make it a little more consistent:

https://github.com/cfengine/modules/blob/master/promise-types/git/README.md

vpodzime:
> use the same format as the other promise types to make it a little more consistent

I don't like the attributes table, especially when attributes here can have multiple various types.

olehermanse:
> I don't like the attributes table, especially when attributes here can have multiple various types.

As a user I like the information it provides, and also seeing somewhat consistent information across promise types is nice.

vpodzime:
> also seeing somewhat consistent information across promise types is nice.

Then we should come up with a better format and make it consistent across the modules.

olehermanse (approved):
LGTM, thanks.

PR #37, opened 2021-11-01
Created README for groups & changed name from groups-experimental to groups

larsewi:


promise-types/groups/README.md
```diff
@@ -0,0 +1,72 @@
+# groups promise type
+
+## Synopsis
+
+* *Name*: `groups`
+* *Version*: `0.1.2`
+* *Description*: Manage local groups.
+
```
olehermanse, line 8: Can you add a note here that this is an experimental version of a promise type, and might be changed in the future?

olehermanse (approved):

olehermanse (approved):

PR #39, opened 2021-11-02
Payload for HTTP(S) promises

vpodzime:


promise-types/http/http_promise_type.py
```diff
@@ -47,10 +47,10 @@ def validate_promise(self, promiser, attributes):
...
                 raise ValidationError("'headers' must be a string, an slist or a data value with 'name: value' pairs")
 
-        if "data" in attributes:
-            data = attributes["data"]
-            if type(data) not in (str, dict):
-                raise ValidationError("'data' must be a string or a data value")
+        if "payload" in attributes:
+            payload = attributes["payload"]
+            if type(payload) not in (str, dict):
+                raise ValidationError("'payload' must be a string or a data value")
```
olehermanse, line 53: I think `data container` is what we usually call this

promise-types/http/http_promise_type.py
```diff
@@ -52,6 +52,9 @@ def validate_promise(self, promiser, attributes):
             if type(payload) not in (str, dict):
                 raise ValidationError("'payload' must be a string or a data value")
 
+            if type(payload) == str and payload.startswith("@") and not os.path.isabs(payload[1:]):
+                raise ValidationError("File-based payload must be an absolute path")
```
olehermanse, line 56: I think I would prefer a more explicit `payload_file` or something similar, instead of a magical `@` sign, but this is entirely subjective, of course.

promise-types/http/http_promise_type.py
```diff
@@ -88,9 +91,22 @@ def evaluate_promise(self, promiser, attributes):
                 if "Content-Type" not in headers:
                     headers["Content-Type"] = "application/json"
 
+            elif payload.startswith("@"):
+                path = payload[1:]
+                try:
+                    # Closed automatically when this variable gets out of
+                    # scope. Thank you, Python!
+                    payload = open(path, "rb")
```
olehermanse, line 99: wait, really? I though that was why we use with / context managers?

olehermanse (approved):
Really nice functionality. I left some small comments, but no blockers.

promise-types/http/http_promise_type.py
vpodzime: I cannot say I disagree. But `@` is quite commonly used to say the given string is actually a file path. Let's see if/how it works for people, we can change it later if desired.

promise-types/http/http_promise_type.py
vpodzime: Yes, you can check it with this, for example:
```
import os
f = open("/etc/passwd")
fd = f.fileno()
print("File points to: " + os.readlink("/proc/%d/fd/%d" % (os.getpid(), fd)))
f = None
print("open" if os.path.exists("/proc/%d/fd/%d" % (os.getpid(), fd)) else "closed")
```

vpodzime: `with` just gives better control of when the file is actually closed.

PR #40, opened 2021-11-02
Result classes for the http promise type

vpodzime:
~Requires https://github.com/cfengine/modules/pull/39 to be merged first. Only the last 3 commits are relevant for this PR.~

### DIFF
--- a/promise-types/http/README.org
+++ b/promise-types/http/README.org
@@ -3,7 +3,7 @@
 ** Synopsis
 
 - /Name/: =http=
-- /Version/: =0.0.3=
+- /Version/: =1.0.0=
 - /Description/: Perform HTTP(S) requests from policy.
 
 ** Promiser and attributes
--- a/promise-types/http/README.org
+++ b/promise-types/http/README.org
@@ -24,6 +24,23 @@ the URL is specified in the promiser.
 | =insecure= | =string=                    | Whether to skip TLS validation of the server's certificate (/"true"/) or not (/"false"/)                                                                         | No        | /"false"/  |
 
 
+** Result classes
+   :PROPERTIES:
+   :CUSTOM_ID: result-classes
+   :END:
+
+On success, each /http/ promise defines a class that is in the form of
+~$(this.promiser)_METHOD_request_done~. On failure, class of form
+~$(this.promiser)_METHOD_request_failed~ is defined plus classes of form:
+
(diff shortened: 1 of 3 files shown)


promise-types/http/README.org
```diff
@@ -24,6 +24,23 @@ the URL is specified in the promiser.
...
+On success, each /http/ promise defines a class that is in the form of
+~$(this.promiser)_METHOD_request_done~. On failure, class of form
+~$(this.promiser)_METHOD_request_failed~ is defined plus classes of form:
+
+- ~$(this.promiser)_METHOD_payload_failed~ (payload-related failures)
+- ~$(this.promiser)_METHOD_payload_file_failed~ (file-based payload failures)
+- ~$(this.promiser)_METHOD_payload_conversion_failed~ (payload conversion failures)
+- ~$(this.promiser)_METHOD_file_failed~ (/file/-related failures, see above)
+
+in the respective cases. *The promiser is canonified in all the above classes.*
```
olehermanse, line 41: What about failures related to the response code? Like a 400 client error, those are the ones I am most interested in.

promise-types/http/http_promise_type.py
```diff
@@ -75,6 +75,8 @@ def evaluate_promise(self, promiser, attributes):
         target = attributes.get("file")
         insecure = attributes.get("insecure", False)
 
+        canonical_promiser = promiser.translate(str.maketrans({char: "_" for char in ("@", "/", ":", "?", "&", "%")}))
```
olehermanse, line 78: We could add this to cfengine.py

promise-types/http/README.org
olehermanse: I see that this will cause the generic one, but not a more specific class. Worth mentioning in the docs.

olehermanse (approved):
LGTM, just some smaller comments.

promise-types/http/http_promise_type.py
vpodzime: Well, generally the promiser can have other various characters, right? And there's actually code on the cf-agent side that makes sure all the result classes are canonified. So this is really just for having nicer logs and such.

olehermanse: I meant we can provide a canonify function that you can use in cases like this.

PR #41, opened 2021-11-02
Added 2 small convenience modules - demo and every-minute

olehermanse:


nickanderson (approved):
I don't approve, but I do.

PR #43, opened 2022-01-25
CFE-3903: Groups promise type now supports custom bodies

larsewi:
Depends on: https://github.com/cfengine/core/pull/4884

 - [x] test on 3.18.x
 - [x] test on master

### DIFF
--- a/promise-types/groups/groups.cf
+++ b/promise-types/groups/groups.cf
@@ -1,7 +1,8 @@
 # Example policy illustraing the use of custom promise module groups.
 #
-# Note: As of 3.18.0, custom promise modules does not support bodies, thus
-# JSON strings are used to define members for now.
+# Note: As of 3.20.0, custom promise modules support the use of custom bodies.
+#       Versions before 3.20.0 uses JSON strings or data containers to define
+#       members instead.
 
 
 promise agent groups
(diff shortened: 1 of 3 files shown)


larsewi:
Tested:
 - [x] 3.18.x
 - [x] 3.20.x

promise-types/groups/groups.py
```diff
@@ -79,9 +79,20 @@ def validate_promise(self, promiser, attributes):
 
         # check attribute members if present
         if "members" in attributes:
-            # parse json
-            members = json.loads(attributes["members"])
-            attributes["members"] = members
+            # pre CFEngine 3.20: we used type data to imitate bodies
```
vpodzime, line 78: `string` containing encoded JSON was used, not `data`.

promise-types/groups/groups.cf
```diff
@@ -36,11 +50,21 @@ bundle agent example_groups
   groups:
     "foo"
       policy => "present",
+@if minimum_version(3.20)
+      members => foo;
```
vpodzime, line 54: `members` can also be of type `data`, right?

promise-types/groups/enable.cf
```diff
@@ -1,6 +1,6 @@
 promise agent groups
 # @brief Define groups promise type
 {
-  path => "$(sys.workdir)/modules/promises/groups.py";
+  path => "$(sys.workdir)/masterfiles/modules/promises/groups.py";
```
vpodzime, line 4: There's no `masterfiles` directory on hosts, it's only present on hubs.

vpodzime (commented):
Looks generally good to me. Some small changes needed.

promise-types/groups/groups.py
```diff
@@ -126,7 +133,9 @@ def evaluate_promise(self, promiser, attributes):
 
         # parse json in attribute members
         if "members" in attributes:
-            attributes["members"] = json.loads(attributes["members"])
+            # if members attribute is passed as a string, parse it as json
+            if type(attributes["members"]) is str:
+                attributes["members"] = json.loads(attributes["members"])
```
olehermanse, line 138: If it's not a string, what is it?

larsewi: If it's not a string, then it must be a dict. Anything else would cause a validation error to be thrown before evaluation.

promise-types/groups/groups.py
larsewi: Yes, you're right. MB

promise-types/groups/groups.cf
larsewi: Yes, I believe so. Do you want me to include an example using the type `data`?

promise-types/groups/enable.cf
larsewi: This should be `"$(sys.workdir)/inputs/modules/promises/groups.py"`?

promise-types/groups/groups.py
larsewi: I can add an assert to make this more clear

promise-types/groups/enable.cf
craigcomstock: according to https://docs.cfengine.com/docs/master/reference-special-variables-sys.html you probably want

`$(sys.inputdir)` # aka `/var/cfengine/inputs`

https://docs.cfengine.com/docs/master/reference-special-variables-sys.html#sys-inputdir

promise-types/groups/groups.cf
```diff
@@ -1,33 +1,47 @@
...
-# JSON strings are used to define members for now.
+# Note: As of 3.20.0, custom promise modules support the use of custom bodies.
+#       Versions before 3.20.0 uses JSON strings to define members instead.
 
 
 promise agent groups
 # @brief Define groups promise type.
 {
-  path => "$(sys.workdir)/modules/promises/groups.py";
+  path => "$(sys.workdir)/masterfiles/modules/promises/groups.py";
```
craigcomstock, line 10: same as above, probably use `$(sys.inputdir)` instead?

promise-types/groups/groups.cf
larsewi: I didn't manage to make it work with data containers. But I'll create a another ticket to make this work.

promise-types/groups/enable.cf
olehermanse: @larsewi @craigcomstock why are you changing this at all? AFAIK it was correct as before. The policy craig worked on puts modules there.

promise-types/groups/groups.cf
olehermanse: masterfiles is responsible for putting modules in `/var/cfengine/modules` I don't know why you are making this change.

promise-types/groups/enable.cf
larsewi: MB

promise-types/groups/groups.cf
larsewi: MB

vpodzime (approved):
Looks good to me.

olehermanse (approved):

craigcomstock (approved):

larsewi:
I will create a PR to update the commit sha in the index

PR #47, opened 2022-08-03
ENT-9047: Various improvements to http promise type

Lex-2008:
Merge together: https://github.com/cfengine/build-index/pull/287
- Inverted logic for downloading files over 512KiB.
- Implemented use of temporary file to prevent partial written file from being used.
- Parent directories leading to promised file are now created automatically.
- Stopped returning REPAIRED when no changes to file were made (now returns KEPT).
- Fixed typo.

### DIFF
--- a/promise-types/http/http_promise_type.py
+++ b/promise-types/http/http_promise_type.py
@@ -154,7 +154,7 @@ class HTTPPromiseModule(PromiseModule):
 
         try:      
             with urllib.request.urlopen(request, context=SSL_context) as url_req:
-                if not (200 <= url_req.status <= 300):
+                if not (200 <= url_req.status < 300):
                     self.log_error("Request for '%s' failed with code %d" % (url, url_req.status))
                     return (Result.NOT_KEPT, ["%s_%s_request_failed" % (canonical_promiser, method)])
                 # TODO: log progress when url_req.headers["Content-length"] > REPORTING_THRESHOLD
--- a/promise-types/http/http_promise_type.py
+++ b/promise-types/http/http_promise_type.py
@@ -152,7 +152,7 @@ class HTTPPromiseModule(PromiseModule):
                 SSL_context = ssl.SSLContext()
                 SSL_context.verify_method = ssl.CERT_NONE
 
-        try:      
+        try:
             with urllib.request.urlopen(request, context=SSL_context) as url_req:
                 if not (200 <= url_req.status < 300):
                     self.log_error("Request for '%s' failed with code %d" % (url, url_req.status))
--- a/promise-types/http/http_promise_type.py
+++ b/promise-types/http/http_promise_type.py
@@ -1,20 +1,27 @@
 """HTTP module for CFEngine"""
 
+import filecmp
 import os
 import urllib
 import urllib.request
 import ssl
 import json
+from contextlib import contextmanager
 
 from cfengine import PromiseModule, ValidationError, Result
 
 
 _SUPPORTED_METHODS = {"GET", "POST", "PUT", "DELETE", "PATCH"}
 
+class FileInfo:
+    def __init__(self, target):
+        self.target = target
+        self.was_repaired = False
+
 
 class HTTPPromiseModule(PromiseModule):
-    def __init__(self, *args, **kwargs):
-        super().__init__("http_promise_module", "1.0.0", *args, **kwargs)
+    def __init__(self, name="http_promise_module", version="1.0.0", **kwargs):
+        super().__init__(name, version, **kwargs)
 
     def validate_promise(self, promiser, attributes):
         if "url" in attributes:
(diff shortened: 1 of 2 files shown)


vpodzime (commented):
Please use `KiB` as the unit in the commit message and mention it's a fix in the *http* promise type.

vpodzime:
And thanks for the fix, of course!

Lex-2008:
All thanks go to reporter in the support ticket! (I don't know if we can disclose their name in a public repo?)

olehermanse:
> All thanks go to reporter in the support ticket! (I don't know if we can disclose their name in a public repo?)

@Lex-2008 definitely not :) (Unless you ask for permission, of course)

nickanderson:
I pushed a change to change the commit per comment from @vpodzime .

Is this ready to go in now?

Side note: Does this module *require* python 3?

vpodzime (commented):
The commit now has many more changes. Ideally, they should gop into separate commits or else the commit message should be amended to mention them.

nickanderson:
> The commit now has many more changes. Ideally, they should gop into separate commits or else the commit message should be amended to mention them.
OK.

@vpodzime I looked up the last comment from the original author of the change:

```
Change `__init__`'s signature so that the `HTTPPromiseModule` can be extended.
Create a new `target_fh` context that simplifies the logic for if target is set or not.
`target_fh` will write to a temp file (this is important so that you don't run into the condition where a client pulls files from the hub while a file is still being written).
`target_fh` creates the path to the target, if it doesn't exist.
`evaluate_promise` returns KEPT if there was not any changes to the file, otherwise the promise is REPAIRED. (before it was always repaired)
```

I used that to formulate a new commit message.

promise-types/http/http_promise_type.py
```diff
@@ -66,6 +73,22 @@ def validate_promise(self, promiser, attributes):
...
+            temp_file = file_info.target+".cftemp"
+            with open(temp_file, "wb") as fh:
+                yield fh
+            if not os.path.isfile(file_info.target) or not filecmp.cmp(temp_file, file_info.target):
+                os.replace(temp_file, file_info.target)
+                file_info.was_repaired = True
+            else:
+                os.remove(temp_file)
+        else:
+            yield open(os.devnull, "wb")
```
craigcomstock, line 90: What is the purpose of returning `os.devnull` aka `/dev/null` here? If I read the code correctly the request would occur, the contents read and written to `/dev/null` with the result would always be `KEPT`. I suppose this is to do something like API requests where you don't care about the result other than response code?

If so, I think a comment here would be helpful to explain and possibly explain in docs the option to provide no `file` attribute as well.

craigcomstock (changes requested):
Looks like this PR addresses two TODOs in the README that should be updated as well as add a note to the README about skipping the "file" attribute as a viable option for some particular purpose.

At least fix the spelling in README `s/Lenght/Length/`

```
[ ] progress reporting if response Content-Lenght is big
[ ] GET requests should not overwrite data if it is the same
```

promise-types/http/http_promise_type.py
```diff
@@ -127,29 +151,21 @@ def evaluate_promise(self, promiser, attributes):
...
-            else:
-                with urllib.request.urlopen(request, context=SSL_context) as url_req:
-                    if not (200 <= url_req.status <= 300):
-                        self.log_error("Request for '%s' failed with code %d" % (url, url_req.status))
-                        return (Result.NOT_KEPT, ["%s_%s_request_failed" % (canonical_promiser, method)])
-                    done = False
-                    while not done:
+        try:      
+            with urllib.request.urlopen(request, context=SSL_context) as url_req:
+                if not (200 <= url_req.status <= 300):
```
craigcomstock, line 156: Interestingly, 300 code (rarely used), means "multiple choices" for redirect so it seems this should be `< 300` instead of `<= 300`.
```suggestion
                if not (200 <= url_req.status < 300):
```

promise-types/http/http_promise_type.py
craigcomstock: @nickanderson I would suggest an addition to the README, maybe near Result Classes of:

```
Note: if no `file` attribute is provided the request will be sent to `/dev/null`. This is useful for something like a REST API endpoint where all you care about is an OK response (http code between 200 and 300) that results in the promise being `KEPT`
```

promise-types/http/http_promise_type.py
nickanderson: Added in separate commit.

craigcomstock (approved):
🍕 !!!

nickanderson:
Will leave merging for someone else in case there are more complaints. Added corollary PR to update build index.

olehermanse (approved):
Looks good to me :)

nickanderson:
Thanks for the contribution @bim9262 !

bim9262:
You're welcome!

PR #56, opened 2022-12-05
Added module for removing .shosts files from home directories

nickanderson:
Ticket: ENT-9662
Changelog: Title

# TODO
- [x] Screenshots
- [x] Example output


cfbs.json
```diff
@@ -87,6 +87,14 @@
         "bundles uninstall_rsh_server",
         "policy_files services/cfbs/modules/uninstall-rsh-server/uninstall-rsh-server.cf"
       ]
+    },
+    "remove-home-dotshosts": {
+      "description": "Policy to make sure that ~/.shosts is not present without an exception.",
```
craigcomstock, line 92: As a description I don't really understand "without an exception" in this context. I imagine you are describing that you can add an exception class but I don't think that belongs in the description.

management/remove-home-dotshosts/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+With Mission Portal you can find hosts which have `~/.shosts` files and details about reasons for exception.
+
+![Inventory Report showing found ~/.shosts and Exception](https://raw.githubusercontent.com/cfengine/modules/master/management/remove-home-dotshots/media/inventory-report.png)
+
+## Example
+
+If you try creating a `~/.shosts` file running the agent with this module, you should see it get removed:
+
+Example output with an exception present:
```
craigcomstock, line 17: I think if you are going to say "exception present" you would have needed to explain what an exception was before-hand. aka the class to not enforce this policy on a particular host.

management/remove-home-dotshosts/README.md
```diff
@@ -0,0 +1,41 @@
...
+
+## Configuration
+
+### Specifying home directory roots to search
+
+Define `remove_home_dotshosts:main.home_dir_roots` as a list or data container array with each element being a home directory root to search.
+
+### Adding exceptions
+
+If `~/.shosts` files are really needed on some hosts, you can add an exception by defining the variable `remove_home_dotshosts.main.exception` containing text describing why the exception is granted and where to find further information. This variable can be defined from the `Host specific data` section in host info pages inside Mission Portal, the CFEngine Enterprise Web UI.
```
craigcomstock, line 41: I don't think that `remove_home_dotshosts.main.exception` is very easy to use. Why not follow the pattern we have so far of `exception_remote_home_dotshosts` arguably no better but at least follows the pattern of "exception_<modulename>".

craigcomstock: https://tracker.mender.io/browse/ENT-9670 (ticket for making module exceptions "easy to use") and move away from these potentially longer and longer class names that must be understood and recognized by a user.

olehermanse (approved):

management/remove-home-dotshosts/README.md
nickanderson: "easy" is only whatever is familiar as far as I am concerned.

Classes have such limited character space, where variable values can be rich and expressive. Classes can have comments (depending on how they are defined), but Mission Portal has no way to make use of those comments where as variables can be displayed.

PR #57, opened 2022-12-05
Added enable-aslr module

craigcomstock:
Ticket: ENT-9660


security/enable-aslr/enable-aslr.cf
```diff
@@ -0,0 +1,42 @@
+bundle agent enable_aslr
+# @brief Makes sure that aslr (Address space layout randomization) is enabled on the system.
```
olehermanse, line 2:
```suggestion
# @brief Makes sure that ASLR (Address Space Layout Randomization) is enabled on the system.
```

security/enable-aslr/README.md
```diff
@@ -0,0 +1,61 @@
+[Address space layout randomization (`aslr`)](https://en.wikipedia.org/wiki/Address_space_layout_randomization) helps prevent exploits by randomly changing the address at which programs are located.
```
olehermanse, line 1:
```suggestion
[Address space layout randomization (`ASLR`)](https://en.wikipedia.org/wiki/Address_space_layout_randomization) helps prevent exploits by randomly changing the address at which programs are located.
```

security/enable-aslr/README.md
```diff
@@ -0,0 +1,61 @@
...
+Q: "...sysctl --system": kernel.random.read_wakeup_threshold = 64
+Q: "...sysctl --system": kernel.random.write_wakeup_threshold = 3072
+Q: "...sysctl --system": * Applying /etc/sysctl.d/60-enable-aslr.conf ...
+Q: "...sysctl --system": kernel.randomize_va_space = 2
+Q: "...sysctl --system": * Applying /etc/sysctl.d/99-sysctl.conf ...
+Q: "...sysctl --system": net.ipv6.conf.all.disable_ipv6 = 1
+Q: "...sysctl --system": * Applying /etc/sysctl.conf ...
+Q: "...sysctl --system": net.ipv6.conf.all.disable_ipv6 = 1
+    info: Last 28 quoted lines were generated by promiser 'sysctl --system'
+    info: Completed execution of 'sysctl --system'
```
olehermanse, line 54: @nickanderson is there a way to make the output a bit nicer here?

security/enable-aslr/enable-aslr.cf
```diff
@@ -0,0 +1,42 @@
...
+      # or `man procfs`
+      # for description of the three modes: 0,1,2
+      "aslr_mode[0]" string => "Disabled";
+      "aslr_mode[1]" string => "Conservative Randomization";
+      "aslr_mode[2]" string => "Full Randomization";
+
+      "aslr_value" string => readfile("/proc/sys/kernel/randomize_va_space");
+      "randomize_va_space_inventory" string => "${aslr_mode[${aslr_value}]} (${aslr_value})",
+        meta => { "inventory", "attribute_name=Address space layout randomization (ASLR)" },
+        comment => "Report on Address space layout randomization (aslr) mode";
```
olehermanse, line 20:
```suggestion
        comment => "Report on Address space layout randomization (ASLR) mode";
```

security/enable-aslr/README.md
```diff
@@ -0,0 +1,61 @@
...
+```
+
+## Adding exceptions
+
+If `aslr` does not need to be enabled for some reason, you can add an exception with the `exception_enable_aslr` class.
```
nickanderson, line 59: I suggest being explicit about the namespace here.

Since your policy is in the default namespace, you probably mean to say `default:exception_enable_aslr`. It's important because if you enter `exception_enable_aslr` in Mission Portal host specific data your going to get a class defined in the `data` namespace.

image

Furthermore, I would actually suggest switching to a *variable* instead of a class for the exception so that you can add useful information about the exception. I suppose that's possible to do via `description` in cmdb which is ultimately comment on the class so certainly arguable against using a variable as I suggest.

security/enable-aslr/enable-aslr.cf
```diff
@@ -0,0 +1,42 @@
...
+    linux::
+      # see https://www.kernel.org/doc/html/latest/admin-guide/sysctl/kernel.html#randomize-va-space
+      # or `man procfs`
+      # for description of the three modes: 0,1,2
+      "aslr_mode[0]" string => "Disabled";
+      "aslr_mode[1]" string => "Conservative Randomization";
+      "aslr_mode[2]" string => "Full Randomization";
+
+      "aslr_value" string => readfile("/proc/sys/kernel/randomize_va_space");
+      "randomize_va_space_inventory" string => "${aslr_mode[${aslr_value}]} (${aslr_value})",
```
nickanderson, line 18: Ah, clever. I was about to ask why you defined the `aslr_mode` associative array above. I like how you used this to get a description in the inventory.

I think that the `aslr_mode` variables are reported back to the hub, perhaps check to see? If you `_` prefix them or add `meta => { "noreport" }` then they should not be reported to the hub.

nickanderson (changes requested):

security/enable-aslr/README.md
craigcomstock:
```
root@ubuntu1604:/var/cfengine/outputs# sysctl --load /etc/sysctl.d/60-enable-aslr.conf
kernel.randomize_va_space = 2
```

nickanderson: you could do `inform => "false"` on the commands promise to suppress the command output.
https://docs.cfengine.com/docs/master/reference-promise-types-commands.html#top

security/enable-aslr/README.md
olehermanse: I think I slightly prefer class + comment. It seems more like the intention for classes to enable / disable things, than `isvariable()`.

security/enable-aslr/enable-aslr.cf
```diff
@@ -0,0 +1,42 @@
...
+
+  files:
+    linux.!aslr_exception_allowed.!aslr_enabled::
+      "/etc/sysctl.d/60-enable-aslr.conf"
+        content => "kernel.randomize_va_space = 2",
+        classes => if_repaired("enable_aslr_repaired");
+
+  commands:
+    enable_aslr_repaired|!aslr_enabled::
+      "sysctl --system"
```
nickanderson, line 40: This ends up loading a lot of stuff that may have had other changes right? Ideally we only change what we need to related to what we are doing.

I saw your comment about systemctl --load ... made me wonder a bit about what happened if you had this enabled and then disabled/excepted after. Probably just need to be clear in the docs that this module does not disable ASLR if it has been previously enabled.

security/enable-aslr/enable-aslr.cf
olehermanse: @craigcomstock did you check "variables" in host info page? (Not inventory attributes)

nickanderson: Cool I have seen some reported without _ prefix but I am often using bundles in non default namespace, so that;s plausibly why,.

craigcomstock: Yes. I checked variables table as well.

security/enable-aslr/enable-aslr.cf
```diff
@@ -0,0 +1,42 @@
...
+    "aslr_exception_allowed"
+      or => {
+        "hardening_aslr_disabled_allowed",
+        "data:hardening_aslr_disabled_allowed",
+        "exception_enable_aslr",
+        "data:exception_enable_aslr",
+      };
+
+  files:
+    linux.!aslr_exception_allowed.!aslr_enabled::
```
olehermanse, line 33:
```suggestion
    !aslr_management_disabled.(linux.!aslr_enabled)::
```

olehermanse (changes requested):
The class guards were quite hard to read for me, so I suggested a new name and order. I chose to highlight the exception part first, meaning to clearly show that these 2 promises should do nothing if it has been disabled by an exception. (The extra parenthesis is intentional in this case).

security/enable-aslr/enable-aslr.cf
```diff
@@ -0,0 +1,42 @@
...
+      "/etc/sysctl.d/60-enable-aslr.conf"
+        content => "kernel.randomize_va_space = 2",
+        classes => if_repaired("enable_aslr_repaired");
+
+  commands:
+    enable_aslr_repaired|!aslr_enabled::
```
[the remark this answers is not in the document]
craigcomstock: removed enable_aslr_repaired (conf file repair) and added linux instead. :+1:

security/enable-aslr/enable-aslr.cf
```diff
@@ -0,0 +1,44 @@
...
+      "randomize_va_space_inventory" string => "${aslr_mode[${aslr_value}]} (${aslr_value})",
+        meta => { "inventory", "attribute_name=Address space layout randomization (ASLR)" },
+        comment => "Report on Address space layout randomization (ASLR) mode";
+
+  classes:
+    "aslr_enabled" expression => strcmp("${aslr_value}", "2");
+    "aslr_management_disabled"
+      or => {
+        "hardening_aslr_disabled_allowed",
+        "data:hardening_aslr_disabled_allowed",
```
nickanderson, line 29: What are the `hardening_aslr_disabled_allowd` classes for?

security/enable-aslr/enable-aslr.cf
```diff
@@ -0,0 +1,44 @@
...
+        "exception_enable_aslr",
+        "data:exception_enable_aslr",
```
nickanderson, line 31: Oh, i see you accounted for Mission Portal class definition here being in the data namesapce. SO I guess the change in the readme isn't strictly needed.

nickanderson (approved):
LGTM

security/enable-aslr/enable-aslr.cf
craigcomstock: removed, was in module I used as a started, I think `disabled_allowed` is confusing anyway.

PR #61, opened 2022-12-06
Add a bit more help to README for delete-home-dotshosts

craigcomstock:
explain "Define" a bit more, as a "variable in augments" and link to both concepts in docs.

### DIFF
--- a/security/delete-home-dotshosts/README.md
+++ b/security/delete-home-dotshosts/README.md
@@ -35,7 +35,8 @@ $ sudo cf-agent -KI
 
 ### Specifying home directory roots to search
 
-In [augments](https://docs.cfengine.com/docs/master/reference-language-concepts-augments.html#top) define a [variable](https://docs.cfengine.com/docs/master/reference-language-concepts-variables.html) `delete_home_dotshosts:main.home_dir_roots` as a list or data container array with each element being a home directory root to search.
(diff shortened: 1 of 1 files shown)


security/delete-home-dotshosts/README.md
```diff
@@ -35,7 +35,7 @@ $ sudo cf-agent -KI
 
 ### Specifying home directory roots to search
 
-Define `delete_home_dotshosts:main.home_dir_roots` as a list or data container array with each element being a home directory root to search.
+In [augments](https://docs.cfengine.com/docs/master/reference-language-concepts-augments.html#top) define a [variable](https://docs.cfengine.com/docs/master/reference-language-concepts-variables.html) `delete_home_dotshosts:main.home_dir_roots` as a list or data container array with each element being a home directory root to search.
```
olehermanse, line 38: I wouldn't phrase it like that, there are many ways to define a variable;

```suggestion
Define a [variable](https://docs.cfengine.com/docs/master/reference-language-concepts-variables.html) `delete_home_dotshosts:main.home_dir_roots` as a string list with each element being a home directory root to search.
One easy way to define a variable is using [augments](https://docs.cfengine.com/docs/master/reference-language-concepts-augments.html#top) (another is using Host specific data, as seen below).
```

craigcomstock: sure. :)

olehermanse (approved):

PR #87, opened 2024-01-08
test ci

oleorhagen:


olehermanse:
@oleorhagen I clicked the button to re-run failed jobs. It now installed cfbs 4.0.2 instead of 4.0.1, and the test no longer fails, so looks like the issue has been fixed 🚀 🎉

PR #88, opened 2024-01-10
Ansible INI module

oleorhagen:
A simple custom promise module wrapping the Ansible INI custom module, and simply passing on attributes from CFEngine into the Ansible module.

The one required attribute for the module is `module_path` which is the path to the `Ansible INI` module installed on the system.

### DIFF
--- a/libraries/python/cfengine.py
+++ b/libraries/python/cfengine.py
@@ -384,6 +384,9 @@ class PromiseModule:
         except Exception as e:
             self.log_critical(
                 "{error_type}: {error}".format(error_type=type(e).__name__, error=e)
+                + e.message
+                if hasattr(e, "message")
+                else ""
             )
             self._add_traceback_to_response()
             self._result = Result.ERROR
--- a/promise-types/ini/README.md
+++ b/promise-types/ini/README.md
@@ -1,0 +1,30 @@
+# Ansible INI promise type
+
+## Synopsis
+
+* *Name*: `Ansible INI - Custom Promise Module`
+* *Version*: `0.0.1`
+* *Description*: Manage configuration files through the Ansible INI module in CFEngine.
+
+## Requirements
+
+* Python installed on the system
+* `ansible` pip package
+* Correct path to the `ini_file.py` in the custom promise module
+
+## Attributes
+
+See [anible_ini module](https://docs.ansible.com/ansible/latest/collections/community/general/ini_file_module.html).
+
+## Example
+
+```cfengine3
+bundle agent main
+{
+  ini:
+    "/path/to/file.ini"
+        section => "foo",
+        option  => "bar",
+        value   => "baz";
+}
+```
--- a/promise-types/ini/example.cf
+++ b/promise-types/ini/example.cf
@@ -1,0 +1,20 @@
+promise agent ini
+# @brief Define ini promise type
+{
+  path        => "$(sys.workdir)/modules/promises/ini.py";
+  interpreter => "/usr/bin/python3";
+}
+
+bundle agent main
+{
+
+  meta:
+    "bundle_version" string => "0.0.1";
+    "promise_type"   string => "ini";
+
+  ini:
+    "/tmp/ini/test.ini"
+      section => "foo",
+      option  => "bar",
+      value   => "baz";
+}
--- a/promise-types/ini/ini.py
+++ b/promise-types/ini/ini.py
@@ -1,0 +1,105 @@
+"""A CFEngine custom promise module for INI files"""
+
+import json
+import subprocess
+import sys
+
+from cfengine import PromiseModule, ValidationError, Result
+
+
+class AnsiballINIModule(PromiseModule):
+    def __init__(self):
+        super().__init__("ansible_ini_promise_module", "0.0.1")
+
+        self.add_attribute("path", str, default_to_promiser=True)
+
+    def validate_attributes(self, promiser, attributes, meta):
+        # Just pass the attributes on transparently to Ansible INI The Ansible
+        # module will report if the missing parameters are not in Ansible attributes
+
+        return True
+
+    def validate_promise(self, promiser: str, attributes: dict, meta: dict) -> None:
+        self.log_error(
+            "Validating the ansible ini promise: %s %s %s"
+            % (promiser, attributes, meta)
+        )
+        if not meta.get("promise_type"):
+            raise ValidationError("Promise type not specified")
+
+        assert meta.get("promise_type") == "ini"
+
+    def evaluate_promise(self, promiser: str, attributes: dict, meta: dict):
+        self.log_error(
+            "Evaluating the ansible ini promise %s, %s, %s"
+            % (promiser, attributes, meta)
+        )
+
+        if "module_path" not in attributes:
+            attributes.setdefault(
+                "module_path",
+                "/tmp/ini_file.py",
+            )
+
+        # NOTE: INI module specific - should not be passed on to Ansible
+        module_path = attributes["module_path"]
+        del attributes["module_path"]
+
+        # NOTE - needed because 'default_to_promiser' is not respected
+        attributes.setdefault("path", promiser)
+
+        self.log_error(
+            "Evaluating the ansible ini promise %s, %s, %s"
+            % (promiser, attributes, meta)
+        )
+
+        proc = subprocess.run(
+            [
+                "python",
+                module_path,
+            ],
+            input=json.dumps({"ANSIBLE_MODULE_ARGS": attributes}).encode("utf-8"),
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
+        )
+
+        if not proc:
+            self.log_error("Failed to run the ansible module")
+            return (
+                Result.NOT_KEPT,
+                [],
+            )
+
+        if proc.returncode != 0:
+            self.log_error("Failed to run the ansible module")
+            self.log_error("Ansible INI module returned(stdout): %s" % proc.stdout)
+            self.log_error("Ansible INI module returned(stderr): %s" % proc.stderr)
+            return (
+                Result.NOT_KEPT,
+                [],
+            )
+
+        self.log_debug("Received output: %s (stdout)" % proc.stdout)
+        self.log_debug("Received output: (stderr): %s" % proc.stderr)
+
+        try:
+            d = json.loads(proc.stdout.decode("UTF-8").strip())
+            if d.get("changed", False):
+                self.log_info(
+                    "Edited content of '%s' (%s)" % (promiser, d.get("msg", ""))
+                )
+        except Exception as e:
+            self.log_error(
+                "Failed to decode the JSON returned from the Ansible INI module. Error: %s"
+                % e
+            )
+            return (Result.NOT_KEPT, [])
+
+        return (
+            Result.KEPT,
+            [],
+        )
+
+
+if __name__ == "__main__":
+    AnsiballINIModule().start()
--- a/cfbs.json
+++ b/cfbs.json
@@ -180,6 +180,15 @@
         "append enable.cf services/init.cf"
       ]
     },
+    "promise-type-ini": {
+      "description": "A Custom CFEngine promise module, installing, and using the Ansible INI module to provide an INI promise type for INI files",
+      "subdirectory": "promise-types/ini",
+      "dependencies": ["library-for-promise-types-in-python"],
+      "steps": [
+        "copy ini.py modules/promises/",
+        "append enable.cf services/init.cf"
+      ]
+    },
     "uninstall-packages": {
       "description": "Allows you to specify a list of packages you want uninstalled on your hosts.",
       "subdirectory": "security/uninstall-packages",
--- a/promise-types/ini/enable.cf
+++ b/promise-types/ini/enable.cf
@@ -1,0 +1,6 @@
+promise agent ini
+# @brief Define ini promise type
+{
+  path        => "$(sys.workdir)/modules/promises/ini.py";
+  interpreter => "/usr/bin/python3";
+}


promise-types/ini/ini.py
```diff
@@ -0,0 +1,107 @@
...
+        self.add_attribute("no_extra_spaces", str, default="yes")
+        self.add_attribute("option", str)
+        self.add_attribute("owner", str)
+        self.add_attribute("section", str, required=True)
+        # TODO - Add seLinux attrs
+        # self.add_attribute("selevel", str, default="sO")
+        # self.add_attribute("serole")
+
+        self.add_attribute("state", str, default="present")
+        self.add_attribute("unsafe_writes", bool, default=False)
```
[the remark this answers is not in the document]
oleorhagen: Good point :p

However, I'm unsure if we should add these at all, or just pass them through transparently?

oleorhagen:
For now I've added a `cfbs.json` file in the directory. However, I see all other modules are incorportated into the top-level `cfbs.json`. I'm guessing you want me to follow that example?

olehermanse:
> For now I've added a `cfbs.json` file in the directory. However, I see all other modules are incorportated into the top-level `cfbs.json`. I'm guessing you want me to follow that example?

@oleorhagen Yea. Either that, or you can also develop the module in your own repo, if you prefer.

promise-types/ini/example.cf
```diff
@@ -0,0 +1,79 @@
+promise agent ini
```
olehermanse, line 1: @oleorhagen sorry that we don't have a good automatic formatter for policy language (yet). Please take a look at the style guide here and try to align with it (doesn't have to be perfect):

https://docs.cfengine.com/docs/master/examples-tutorials-writing-and-serving-policy-policy-style.html#top

promise-types/ini/ini.py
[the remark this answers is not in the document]
oleorhagen: Ended up just passing them through transparently

promise-types/ini/example.cf
oleorhagen: Did my best. See latest push

promise-types/ini/README.md
```diff
@@ -0,0 +1,28 @@
...
+
+## Synopsis
+
+* *Name*: `Ansible INI - Custom Promise Module`
+* *Version*: `0.0.1`
+* *Description*: Manage TOML configuration files through the Ansible INI module in CFEngine.
+
+## Requirements
+
+* Python installed on the system (?)
```
oleorhagen, line 11: Should I leave this in or not?

oleorhagen: Left it in and added the `ansible-core` package requirement as well, also removed them from the policy for now, I don't see any of the other promises installing their dependencies. So trying to keep it consistent.

promise-types/ini/example.cf
```diff
@@ -0,0 +1,69 @@
...
+   "url.max_content": 1048576,
+   "url.verbose": 0
+} ';
+    "options" data => parsejson($(options_str));
+    "url" string => "https://raw.githubusercontent.com/ansible-collections/community.general/9946f758af5fe0fe41f1cf7584d670ca1d6c2d52/plugins/modules/ini_file.py";
+    "ansible_ini_file" data => url_get($(url), options);
+
+  classes:
+
+    "got_ini_file" expression => "$(ansible_ini_file[success])";
```
craigcomstock, line 32: I would expect more of a `have_ansible_ini_file_module` class which hinges on the file being present, not the result of `url_get`.

Best to think of the end-state you wish, not the momentary state. e.g. have-ini-file instead of just-got-ini-file-with-url-get-function. Thinking policy is an imperative language here induces madness and frustration. ;) 

Would be kind of cool if we could get `cfbs build` to do some of this prep work, like download ini_file.py into the policy set, maybe with a checksum. :)

promise-types/ini/example.cf
```diff
@@ -0,0 +1,69 @@
...
+
+
+  files:
+
+    got_ini_file::
+      "/tmp/cfe/ini_file.py"
```
craigcomstock, line 38: for re-use sake I would say put this in the statedir (typically /var/cfengine/state) and set a var for it.

```suggestion
      "${ansible_ini_file_module_path}"
```

so earlier

```
  vars:
    "ansible_ini_file_module_path" string => "${sys.statedir}/ansible_modules/ini_file.py";
```

see https://docs.cfengine.com/docs/3.21/reference-special-variables-sys.html#sys-statedir

promise-types/ini/ini.py
```diff
@@ -0,0 +1,98 @@
...
+        if proc.returncode != 0:
+            self.log_error("Failed to run the ansible module")
+            self.log_error("Ansible INI module returned(stdout): %s" % proc.stdout)
+            self.log_error("Ansible INI module returned(stderr): %s" % proc.stderr)
+            return (
+                Result.NOT_KEPT,
+                [],
+            )
+
+        self.log_debug("Received output: %s (stdout)" % proc.stdout)
```
craigcomstock, line 77: right, here, when returncode is 0 you should parse results somehow and provide some info logging about what changed if anything: file created, file modified, ... maybe by changing the options sent to the ansible module.

craigcomstock (commented):
cool! thanks for this work!

promise-types/ini/example.cf
```diff
@@ -0,0 +1,69 @@
...
+bundle agent main
+{
+
+  meta:
+    "bundle_version" string => "0.0.1";
+    "promise_type" string => "ini";
+
+  ini:
+    "/tmp/ini/test.toml"
+      module_path => "/tmp/cfe/ini_file.py",
```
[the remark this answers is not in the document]
nickanderson: Yeah, I don't think we should have to specify `module_path` on the individual promise. That should be handled by setting up the promise type.

nickanderson (changes requested):

promise-types/ini/ini.py
```diff
@@ -0,0 +1,98 @@
...
+
+    def validate_promise(self, promiser: str, attributes: dict, meta: dict) -> None:
+        self.log_debug(
+            "Validating the ansible ini promise: %s %s %s"
+            % (promiser, attributes, meta)
+        )
+        if not meta.get("promise_type"):
+            raise ValidationError("Promise type not specified")
+
+        assert meta.get("promise_type") == "ini"
```
nickanderson, line 36:
```suggestion
        assert meta.get("promise_type") == "ansible_ini"
```

promise-types/ini/README.md
```diff
@@ -0,0 +1,29 @@
+# Ansible INI promise type
+
+## Synopsis
+
+* *Name*: `Ansible INI - Custom Promise Module`
+* *Version*: `0.0.1`
+* *Description*: Manage TOML configuration files through the Ansible INI module in CFEngine.
```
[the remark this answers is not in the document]
olehermanse: @oleorhagen AFAICT TOML and ini are 2 similar but different file formats. Does the Ansible module work for / target both TOML and ini files?

https://en.wikipedia.org/wiki/INI_file

promise-types/ini/README.md
```diff
@@ -0,0 +1,29 @@
...
+## Attributes
+
+See [anible_ini module](https://docs.ansible.com/ansible/latest/collections/community/general/ini_file_module.html).
+
+## Example
+
+```cfengine3
+bundle agent main
+{
+  ini:
```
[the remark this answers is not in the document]
olehermanse: I don't agree. The promise type should refer to what resource we are managing, not how it was implemented (python, ansible, etc.).

`ini` or maybe `ini_files` (similar to `files` promises) are appropriate names here, IMO.

promise-types/ini/README.md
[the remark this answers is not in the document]
oleorhagen: hah, lol. TIL. Ur right!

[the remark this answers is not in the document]
oleorhagen: Fixing

promise-types/ini/example.cf
oleorhagen: Yes, that is much better!

promise-types/ini/ini.py
oleorhagen: Isn't this what is done below in the `try` ? :point_down:

craigcomstock: I think @olehermanse was trying to mention this. Typically when CFEngine makes a change it logs an INFO message about what it did. I'm not sure your try/catch. I think the message should be more direct and you should build more intelligence into the python code.

It would seem that `changed` key in `d.get("changed")` should log a message like 

```
self.log_info("Ansible INI modified file %s", ...)
```

promise-types/ini/README.md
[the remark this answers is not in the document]
oleorhagen: I'm sticking to `ini` then I think

promise-types/ini/ini.py
```diff
@@ -0,0 +1,98 @@
...
+                [],
+            )
+
+        self.log_debug("Received output: %s (stdout)" % proc.stdout)
+        self.log_debug("Received output: (stderr): %s" % proc.stderr)
+
+        try:
+            d = json.loads(proc.stdout.decode("UTF-8").strip())
+            if d.get("changed", False):
+                self.log_info("Ansible INI returned msg: %s" % d.get("msg", ""))
```
olehermanse, line 83:
```suggestion
                self.log_info("Ansible INI returned msg: %s" % d.get("msg", "empty"))
```

olehermanse:
```
Edited content of 'filename.ini' (msg)
```

oleorhagen:
@olehermanse feel free to have a last lookover

libraries/python/cfengine.py
```diff
@@ -384,6 +384,9 @@ def _handle_evaluate(self, promiser, attributes, request):
         except Exception as e:
             self.log_critical(
                 "{error_type}: {error}".format(error_type=type(e).__name__, error=e)
+                + e.message
```
craigcomstock, line 387: seems like this would read easier and fit in one line.

craigcomstock: then again, what does `black` say? yeah, black seems to agree with me. Maybe run all your python files through `black`

promise-types/ini/README.md
```diff
@@ -0,0 +1,30 @@
...
+    "/path/to/file.ini"
+        section => "foo",
+        option     => "bar",
+        value => "baz";
```
craigcomstock, line 28: either

```suggestion
        section => "foo",
        option => "bar",
        value => "baz";
```

or

```suggestion
        section => "foo",
        option  => "bar",
        value   => "baz";
```

promise-types/ini/example.cf
```diff
@@ -0,0 +1,20 @@
...
+
+  meta:
+    "bundle_version" string => "0.0.1";
+    "promise_type" string => "ini";
+
+  ini:
+    "/tmp/ini/test.ini"
+      section => "foo",
+      option     => "bar",
+      value => "baz";
```
craigcomstock, line 19: same deal here as in other example, tidy up whitespace either tight (one space between attribute name and fat arrow) or lined up vertically.

promise-types/ini/ini.py
```diff
@@ -0,0 +1,109 @@
+"""The ultimate CFEngine custom promise module"""
```
craigcomstock, line 1: 🏆

promise-types/ini/ini.py
```diff
@@ -0,0 +1,109 @@
...
+
+        assert meta.get("promise_type") == "ini"
+
+    def evaluate_promise(self, promiser: str, attributes: dict, meta: dict):
+        self.log_error(
+            "Evaluating the ansible ini promise %s, %s, %s"
+            % (promiser, attributes, meta)
+        )
+
+        if "module_path" not in attributes:
```
craigcomstock, line 42: Related to the TODO above where validation "does not work" here we provide a reasonable default? Doesn't seem like it actually since we are assuming here python3.11 which is a wild assumption for many different platforms.

Can we be smarter here about the path?

I had to `pip3 install ansible` in order to get ini_file.py. `pip3 install ansible-core` didn't do enough.

If I do `pip3 show ansible` I get an idea of where to look that is dependent on the install which might be better and something we can do programmatically via python?

```
? pip3 show ansible
Name: ansible
Version: 9.2.0
Summary: Radically simple IT automation
Home-page: https://ansible.com/
Author: Ansible, Inc.
Author-email: info@ansible.com
License: GPL-3.0-or-later
Location: /usr/local/lib/python3.11/site-packages
Requires: ansible-core
Required-by: 
```

craigcomstock (changes requested):
I think we must make the module_path default more dynamic based on platform/host information.

libraries/python/cfengine.py
oleorhagen: Hmm, this is `black`. But I'm guessing it's a version thing then. I used `19`. but don't know what your standard is

promise-types/ini/ini.py
oleorhagen: I get your point, and I agree. However, I won't have time to finish this, and it seems the other modules too write in their `Readme`'s that the prerequisites need to be installed. 

I fiddled around with how to provide this, and tried a few different things:

1. Having the path in the `promise agent ini` setup, but this apparently is not valid syntax
2. Having it in the `meta` section in the promise - but this does not get passed to the custom promise module (only 
3. Hardcoding it as I've done now.

Dynamically would have been the sweetest though

oleorhagen: > I had to pip3 install ansible in order to get ini_file.py. pip3 install ansible-core didn't do enough.

This is true, if we do not provide the file ourselves (Like in the original implementation where it was downloaded).

we could consider adding the file to the repo, and have it installed just as we do the custom promise module `ini.py` itself?

craigcomstock: Right on. Two things then: 
- we need to update the requirements to include ansible since that's easier than having folks have to figure out where to get it.
- make the path dynamic with some python code if possible, doesn't have to be done by you :)

promise-types/ini/example.cf
```diff
@@ -0,0 +1,20 @@
...
+  path        => "$(sys.workdir)/modules/promises/ini.py";
+  interpreter => "/usr/bin/python3";
+}
+
+bundle agent main
+{
+
+  meta:
+    "bundle_version" string => "0.0.1";
+    "promise_type"   string => "ini";
```
olehermanse, line 13: Using the `meta` promise type won't do anything for the promise module.

You could use the `meta` attribute per promise:

```
  ini:
    "hello.ini"
      meta => { "report", "some=value" };
```

But it seems to me that the information you want to pass is more static, so would fit better if we make some room for it in the `promise agent` block.

promise-types/ini/example.cf
```diff
@@ -0,0 +1,20 @@
+promise agent ini
+# @brief Define ini promise type
+{
+  path        => "$(sys.workdir)/modules/promises/ini.py";
+  interpreter => "/usr/bin/python3";
```
olehermanse, line 5: We could enable setting environment variables for the module, or add another field of "data" to pass to the module here.

oleorhagen: Yes, I like the `data` field idea

PR #89, opened 2024-01-18
CFE-4159: Changed to info log level on promise repaired for `groups` promise type

larsewi:


promise-types/groups/groups.py
```diff
@@ -200,7 +200,7 @@ def evaluate_promise(self, promiser, attributes, meta):
             return Result.NOT_KEPT
 
         if repairs > 0:
-            self.log_verbose("Promise '%s' repaired" % promiser)
+            self.log_info("Promise '%s' repaired" % promiser)
```
olehermanse, line 203: I don't think this is right. The outcome of the promise fits well in verbose log. What we'd like in the info log is real (human readable) information about what was changed.

I see there is already 1 good log message for this above (`Deleted group '%s'`).

So I'd say keep this one as is, but introduce other info log messages further up / where the changes are actually made.

Some ideas;

```
Added 'larsewi' to 'developers' group
Removed 'larsewi' from 'developers' group
Set members of 'developers' group to be: larsewi, craigcomstock 
```

olehermanse (changes requested):

promise-types/groups/groups.py
craigcomstock: Agreed, as we were just discussing this in regards to the ini module. :+1:

larsewi:
Failed to reproduce the error message, even when using the same policy that I used when I logged this ticket. Closing the ticket. Looking through the code, I cannot see any place where the module returns `Result.REPAIRED` without logging a message with log level info.

PR #90, opened 2024-02-28
Renamed meta to metadata and other cleanups

olehermanse:
- Python promise modules: Renamed meta parameter to metadata
- Cleaned up README files relating to the updated promise types
- iptables.py: Autoformatting by black
- Bumped version numbers of to be released modules inside python source code

### DIFF
--- a/libraries/python/README.md
+++ b/libraries/python/README.md
@@ -1,0 +1,16 @@
+A library for CFEngine custom promise types implemented in Python.
+It provides the `PromiseModule` base class and many helper values and mechanisms that facilitates creation of custom promises.
+
+If you'd like to develop new promise types (or other modules) for CFEngine, we recommend finishing our getting started tutorial:
+
+https://docs.cfengine.com/docs/3.21/getting-started.html
+
+(The last part shows implementing your first module).
+
+If you feel comfortable with using CFEngine and writing Python code already, there is a template project you can just copy and start editing:
+
+https://github.com/cfengine/promise-type-template
+
+For more detailed information about how custom promise types work, see our documentation:
+
+https://docs.cfengine.com/docs/3.21/reference-promise-types-custom.html
--- a/promise-types/http/README.org
+++ b/promise-types/http/README.org
@@ -1,22 +1,24 @@
-* http promise type
+The =http= promise type lets you perform HTTP(S) requests from policy.
 
-** Synopsis
+** Example
 
-- /Name/: =http=
-- /Version/: =1.0.0=
-- /Description/: Perform HTTP(S) requests from policy.
+Here is a simple example of downloading an SVG file (with a HTTP GET request):
 
-** Promiser and attributes
-
-The promiser can either be a URL of the request or, if the /url/ attribute is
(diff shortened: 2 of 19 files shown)


libraries/python/README.md
```diff
@@ -0,0 +1,16 @@
+A library for CFEngine custom promise types implemented in Python.
+It provides the `PromiseModule` base class and many helper values and mechanisms that facilitates creation of custom promises.
+
+If you'd like to develop new promise types (or other modules) for CFEngine, we recommend finishing our getting started tutorial:
+
+https://docs.cfengine.com/docs/3.21/getting-started.html
```
nickanderson, line 6:
```suggestion
https://docs.cfengine.com/latest/getting-started.html
```

Maybe link to latest, lts or docs/master ?

libraries/python/README.md
```diff
@@ -0,0 +1,16 @@
...
+
+(The last part shows implementing your first module).
+
+If you feel comfortable with using CFEngine and writing Python code already, there is a template project you can just copy and start editing:
+
+https://github.com/cfengine/promise-type-template
+
+For more detailed information about how custom promise types work, see our documentation:
+
+https://docs.cfengine.com/docs/3.21/reference-promise-types-custom.html
```
nickanderson, line 16: Same here, maybe link to a not specific version

promise-types/http/README.org
```diff
@@ -1,22 +1,24 @@
-* http promise type
+The `http` promise type lets you perform HTTP(S) requests from policy.
```
nickanderson, line 1:
```suggestion
The =http= promise type lets you perform HTTP(S) requests from policy.
```
This one is org-mode formatted. So you would want to use `=`

nickanderson (changes requested):
Seems pretty ok, just saw one case of markdown formatting used in org formatted document that needs to be fixed. Suggested changing links to not version specific.

promise-types/http/README.org
olehermanse: Fixed, thanks.

libraries/python/README.md
nickanderson: yeah i always struggle to decide.

nickanderson (approved):

promise-types/iptables/iptables.py
```diff
@@ -182,7 +182,7 @@ def validate_promise(self, promiser: str, attributes: dict, meta: dict):
         if command != "flush" and attributes.get("chain") == "ALL":
             raise ValidationError("Chain 'ALL' is only available for command 'flush'")
 
-    def evaluate_promise(self, promiser: str, attributes: dict, meta: dict):
+    def evaluate_promise(self, promiser: str, attributes: Dict, metadata: Dict):
```
larsewi, line 185: What is the reason for using `typing.Dict` instead of `dict`?

olehermanse: I didn't mean anything special by it, I just copy-pasted the function signature from one of the other ones (git promise type I think).

It could make sense to always use the typing ones, make everything look consistent, and they have some more expressive ones. But I don't really have a strong opinion.

PR #95, opened 2024-03-07
Added new module to disable automatic key trust

olehermanse:


management/disable-automatic-key-trust/def.json
```diff
@@ -0,0 +1 @@
+{ "variables": {"default:def.trustkeysfrom": []} }
```
olehermanse, line 1: We should improve cfbs pretty;
https://northerntech.atlassian.net/browse/CFE-3785

nickanderson (approved):
Nice

larsewi (approved):

management/disable-automatic-key-trust/README.md
```diff
@@ -0,0 +1,91 @@
...
+{
+  "variables": {
+    "default:def.acl": ["1.2.3.4", "4.3.2.1"],
+    "default:def.control_server_allowconnects": ["1.2.3.4", "4.3.2.1"],
+    "default:def.trustkeysfrom": ["1.2.3.4", "4.3.2.1"]
+  }
+}
+```
+
+Finally, if you want those only those 2 hosts to communicate, and only with those keys they already have, you can do this:
```
olehermanse, line 82:
```suggestion
Finally, if you want only those 2 hosts to communicate, and only with those keys they already have, you can do this:
```

PR #96, opened 2024-04-04
ENT-11525: Added module to disable recommendations from the MPF

nickanderson:
Ticket: ENT-11525
Changelog: Title

You can test this by adding my fork at the commit of the latest change, e.g.:

```
cfbs add https://github.com/nickanderson/modules@09b0a68d6cc16190f348878284a83bcfcddfffc0 disable-recommendations
```

Do we want this? I dunno, maybe it's better as a suggestion in MP rather than being a module. But we have other similar modules and I was just looking to improve quality of life for Vrata.

### DIFF
--- a/management/disable-recommendations/README.md
+++ b/management/disable-recommendations/README.md
@@ -1,0 +1,20 @@
+The Masterfiles Policy Framework (MPF) emits recommendations for various settings given the context and role of a host.
+
+For example, when federated reporting is enabled on an Enterprise hub, there is a recommendation to install gnu-parallel.
+
+```
+R: CFEngine recommends installing gnu parallel on federated reporting superhubs.
+```
+
+These can be useful, but you may want to disable the functionality to quiet your policy runs. This module facilitates disabling all recommendations by defining `default:cfengine_recommendations_disabled`, it is equivalent to editing the augments file (`/var/cfengine/masterfiles/def.json`) to:
+
+```json
+{
+  "classes": {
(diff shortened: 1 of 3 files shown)


management/disable-recommendations/README.md
```diff
@@ -0,0 +1,20 @@
+The Masterfiles Policy Framework (MPF) emits recommendations for various settings given the context and role of a host.
+
+For example, when federated reporting is enabled on an Enterprise hub, there is a recommendation to install gnu-parallel.
+
+```
+R: CFEngine recommends installing gnu parallel on federated reporting superhubs.
+```
+
+These can be useful, but you may want to disable the functionality to quiet your policy runs. This module facilitates disabling all recommendations by defining `default:cfengine_recommendations_disabled`, it is equivalent to editing the augments file (`/var/cfengine/masterfiles/def.json`) to:
```
vpodzime, line 9:
```suggestion
These can be useful, but you may want to disable the functionality to quiet your policy runs. This module facilitates disabling all recommendations by defining the `default:cfengine_recommendations_disabled` class. Thus it is an equivalent to editing the augments file (`/var/cfengine/masterfiles/def.json`) to:
```

vpodzime (approved):
Looks good to me otherwise. Like I wrote somewhere else before, I think it would make more sense to have this as an input/parameter of the MPF module. But not a strong opinion.

olehermanse (approved):

PR #98, opened 2024-06-19
Always send important log messages from Python-based promise types 

vpodzime:
These messages inform about changes being done for repaired
promises or errors reported for failed promises, etc. and the
agent requires some whenever there's repaired/failed promise by
a custom promise type module.

Ticket: CFE-4406
Changelog: None

### DIFF
--- a/libraries/python/cfengine.py
+++ b/libraries/python/cfengine.py
@@ -44,7 +44,13 @@ def _would_log(level_set, msg_level):
         # uknown level, assume it would be logged
         return True
 
-    return _LOG_LEVELS[msg_level] <= _LOG_LEVELS[level_set]
+    # info: log messages are special because they report changes done in promise
+    # evaluation which is important not only for showing to the user, but also
+    # for auditing/changelog and all modules are required to send info: messages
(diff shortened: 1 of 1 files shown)


nickanderson (approved):

libraries/python/cfengine.py
```diff
@@ -44,6 +44,13 @@ def _would_log(level_set, msg_level):
         # uknown level, assume it would be logged
         return True
 
+    # info: log messages are special because they report changes done in promise
+    # evaluation which is important not only for showing to the user, but also
+    # for auditing/changelog and all modules are required to send info: messages
+    # for all REPAIRED promises.
+    if msg_level == "info":
```
olehermanse, line 51: @vpodzime so it needs to be something like:
```suggestion
    if msg_level in ("info", "error", "notice", "critical", "warning"):
```

Also, `_would_log` doesn't fit this. I think you should either move this check to the `_log` function, or change the name of this function from `_would_log` to `_should_send_log_message` or something similar.

olehermanse (approved):

Issue #99, opened 2024-08-06
git promise type crashes with unicode error


basvandervlies:
Files that have data that is not `utf-8` will crash the promise type:
```
   debug: Received line from module: 'log_verbose=Run: git diff ..origin/master'                                                                                                     
 verbose: Run: git diff ..origin/master                                                                                                                                              
   debug: Received line from module: 'log_critical=UnicodeDecodeError: 'utf-8' codec can't decode byte 0xb4 in position 2025: invalid start byte'                                    
CRITICAL: UnicodeDecodeError: 'utf-8' codec can't decode byte 0xb4 in position 2025: invalid start byte                                                                              
   debug: Received line from module: '{"operation": "evaluate_promise", "promiser": "cf_data_repo", "attributes": {"destination": "/data/cfengine3", "force": true, "repository": "git@git.ia.surfsara.nl:clustercomputing/cf_data", "ssh_options": "-i /root/.ssh/id_ed25519 -o IdentitiesOnly=yes", "update": true, "version": "master"}, "log": [{"level": "debug", "message": "Traceback (most recent call last):\n  File \"/var/cfengine/modules/promises/cfengine.py\", line 381, in _handle_evaluate\n    results = self.evaluate_promise(promiser, attributes, metadata)\n              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^\n  File \"/var/cfengine/modules/promises/git.py\", line 190, in evaluate_promise\n    output = self._git(\n             ^^^^^^^^^^\n  File \"/var/cfengine/modules/promises/git.py\", line 247, in _git\n    .decode(\"utf-8\")\n     ^^^^^^^^^^^^^^^\nUnicodeDecodeError: 'utf-8' codec can't decode byte 0xb4 in position 2025: invalid start byte\n"}], "result": "error"}'                                                                                          
   debug: Traceback (most recent call last):                                                                                                                                         
  File "/var/cfengine/modules/promises/cfengine.py", line 381, in _handle_evaluate                                                                                                   
    results = self.evaluate_promise(promiser, attributes, metadata)                                                                                                                  
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^                                                                                                                  
  File "/var/cfengine/modules/promises/git.py", line 190, in evaluate_promise                                                                                                        
    output = self._git(                                                                                                                                                              
[...]

PR #100, opened 2024-08-06
git promise type must just continue if we have unicode error

basvandervlies:
closes https://github.com/cfengine/modules/issues/99

### DIFF
--- a/promise-types/git/git.py
+++ b/promise-types/git/git.py
@@ -192,6 +192,7 @@ class GitPromiseTypeModule(PromiseModule):
                             [
                                 model.executable,
                                 "diff",
+                                "--shortstat",
                                 "..{remote}/{version}".format(
                                     remote=model.remote, version=model.version
                                 ),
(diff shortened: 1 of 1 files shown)


olehermanse:
@basvandervlies Thanks. It seems this will work at least for this case since the command being run is `git diff` and what we do with it is check if the output is empty. However, for other uses of the `_git` function, I think your solution could be problematic. How about taking the output, attempting to decode to `utf-8`, if that fails, fall back to filtered ascii? I.e. we can loop through the array and replace anything which is not ascii with `?` mark. That seems to mostly preserve the output, to make a nicer log message, and to preserve the emptyness / length of the output as well.

basvandervlies:
@olehermanse yes I noticed it to check if there is output. This output can be huge and for me it was a surprise that is crashes with an `unicode` error. I think this will only occurs  with the `git diff`  command. The second thing is are we interested in the `diff` output? It makes more sense to do this:
```
 13:41 mgmt1:/data/cfengine3 (master)
root# git diff  --shortstat ..origin/master

13:41 mgmt1:/data/cfengine3 (master)
root# git diff  --shortstat ..origin/revert-c7019895
 501 files changed, 4483 insertions(+), 18101 deletions(-)
```

Another thing is why do we not check the exit code of the command to report when it failed or is this done differently?

basvandervlies:
@olehermanse yes I noticed it to check if there is output. This output can be huge and for me it was a surprise that is crashes with an `unicode` error. I think this will only occurs  with the `git diff`  command. The second thing is are we interested in the `diff` output? It makes more sense to do this: (or `--numstat`)
```
 13:41 mgmt1:/data/cfengine3 (master)
root# git diff  --shortstat ..origin/master

13:41 mgmt1:/data/cfengine3 (master)
root# git diff  --shortstat ..origin/revert-c7019895
 501 files changed, 4483 insertions(+), 18101 deletions(-)
```

Another thing is why do we not check the exit code of the command to report when it failed or is this done differently?

basvandervlies:
Some feedback will be nice!! We have this patch in our prod environment and no issues at all.

olehermanse (approved):
ACK. Your improvement makes sense, and fixes the issue, I was just thinking that these functions and their handling of exit code and output is a bit "iffy" (i.e. might cause problem if more git commands are used in the future), but you did not introduce these issues.

olehermanse:
Updating promise-type-git to version 0.2.5 here: https://github.com/cfengine/build-index/pull/516

olehermanse:
@basvandervlies Thank you for your contribution, it has been released as version 0.2.5 of the module: https://build.cfengine.com/modules/promise-type-git/0.2.5/

PR #101, opened 2024-08-30
Added package-module-winget and needed dependency powershell-execution-policy

craigcomstock:
Ticket: ENT-12127

### DIFF
--- a/cfbs.json
+++ b/cfbs.json
@@ -154,6 +154,52 @@
         "bundles maintainers_in_motd"
       ]
     },
+    "powershell-execution-policy": {
+      "description": "Inventory and bundle for PowerShell Execution Policy",
+      "subdirectory": "management/powershell-execution-policy",
+      "steps": [
+        "directory ./ services/cfbs/powershell-execution-policy/",
+        "policy_files services/cfbs/powershell-execution-policy/",
+        "bundles powershell_execution_policy_inventory"
+      ]
+    },
+    "package-method-winget": {
+      "description": "Package method for Windows winget package manager.",
+      "subdirectory": "management/package-method-winget",
+      "dependencies": ["powershell-execution-policy"],
+      "steps": [
+        "input ./input.json def.json",
+        "directory ./ services/cfbs/modules/package-method-winget/",
+        "policy_files services/cfbs/modules/package-method-winget/",
+        "bundles package_method_winget:package_method_winget winget_installed:winget_installed"
+      ],
+      "input": [
+        {
+          "type": "string",
+          "variable": "accept_source_agreements",
+          "namespace": "data",
+          "bundle": "package_method_winget",
+          "label": "Accept Source Agreements",
+          "question": "Would you like to accept source agreements for winget packages promises? [yes|no]"
+        },
+        {
+          "type": "string",
+          "variable": "accept_package_agreements",
+          "namespace": "data",
+          "bundle": "package_method_winget",
+          "label": "Accept Package Agreements",
+          "question": "Would you like to accept package agreements for winget packages promises? [yes|no]"
+        },
+        {
+          "type": "string",
+          "variable": "allow_powershell_execution_policy_change",
+          "namespace": "data",
+          "bundle": "winget_installed",
+          "label": "Allow necessary PowerShell Execution Policy change: LocalMachine set to Unrestricted in order to install winget and cmdlets",
+          "question": "Would you like to allow this module to change PowerShell Execution Policy to LocalMachine:Unrestricted? [yes|no]"
+        }
+      ]
+    },
     "promise-type-ansible": {
       "description": "Promise type to manage systemd services.",
       "subdirectory": "promise-types/ansible",
--- a/management/package-method-winget/README.md
+++ b/management/package-method-winget/README.md
@@ -1,0 +1,52 @@
+# package-method-winget
+
+## Usage
+This module enables a `winget` package method used with policy like:
+
+```cf3
+packages:
+  windows::
+    "Microsoft.WindowsTerminal"
+      package_method => winget,
+      package_policy => "add";
+```
+
+## Opt-in for accepting source and package agreements (Required)
+
+In order for winget to operate properly you must opt-in to accepting source and packaging agreements.
+Without this acceptance the package method will fail with a message like:
+
+```console
+    info: Installing Microsoft.WindowsTerminal...
+    info: Q:powershell.exe -Comm ...:You must set some vars for package-method-winget to work
+   error: Finished command related to promiser 'Microsoft.WindowsTerminal' -- an error occurred, returned 1
+   error: Bulk package schedule execution failed somewhere - unknown outcome for 'Microsoft.WindowsTerminal'
+```
+
+Acceptance can be given either via cfbs inputs, augments, group or host specific data.
+
+Set the value of `yes` in the following variables:
+- `data:package_method_winget.accept_source_agreements`
+- `data:package_method_winget.accept_package_agreements`
+
+## Installation
+This module uses both the `winget` command as well as the `Microsoft.WinGet.Client` PowerShell module as that makes it easier to gather the list of currently installed packages.
+
+`winget` should be installed on most newer desktop systems by default.
+Server images often do not have `winget` installed.
+
+In order for `winget` and `Microsoft.WinGet.Client` to be installed, ps1 scripts must be run which requires `PowerShell Execution policy` `Unrestricted` for `LocalMachine`.
+The policy by default will not make this change.
+You must opt-in by setting the variable `winget_installed.allow_powershell_execution_policy_change` to the value of `yes`.
+This can be set in Host/Group data or via Augments in the `data` namespace.
+
+```json
+{
+  "variables": {
+    "data:winget_installed.allow_powershell_execution_policy_change": {
+      "value": "yes"
+    }
+  }
+}
+```
+
--- a/management/package-method-winget/package-method-winget.cf
+++ b/management/package-method-winget/package-method-winget.cf
@@ -1,0 +1,55 @@
+# NOTE: CASE MATTERS in package names, e.g. Docker.DockerDesktop will match but docker.dockerdesktop won't and so will re-install each agent run!
+# TODO: fix case sensitive package names? ^^^
+body package_method winget
+{
+      package_changes => "bulk";
+      package_name_convention   => "$(name)";
+      package_delete_convention => "$(name)";
+
+      package_installed_regex => ".*";
+      # Note that package_list_name_regex does not allow for commas inside of package names as parsing that from ConvertTo-Csv below would be too complex
+      # an example of output of the package_list_command below is
+      #
+      # "Microsoft.WindowsTerminal","1.21.2361.0"
+      #
+      # so the matches below simply grab the thing before or after the comma separating Id and InstalledVersion fields.
+      package_list_name_regex => '^"([^"]*)",.*';
+      package_list_version_regex => '.*,"([^"]*)".*';
+
+
+      # Here we use the Get-WinGetPackage Cmdlet because it is easier to produce easily parsed information that way
+      package_list_command   => "$(sys.winsysdir)\\WindowsPowerShell\\v1.0\\powershell.exe -Command \"Get-WinGetPackage | Select Id,InstalledVersion | ConvertTo-Csv ";
+      package_delete_command => "$(sys.winsysdir)\\WindowsPowerShell\\v1.0\\powershell.exe -Command \"Uninstall-WinGetPackage ";
+
+      # Here we use winget instead of PowerShell Cmdlets because we can provide the --accept-source-agreements and --accept-package-agreements this way which gets around dialog prompts
+      package_method_winget:accept_source_agreements.package_method_winget:accept_package_agreements::
+        package_add_command    => "$(sys.winsysdir)\\WindowsPowerShell\\v1.0\\powershell.exe -Command \"winget install --accept-source-agreements --accept-package-agreements ";
+      !package_method_winget:accept_source_agreements|!package_method_winget:accept_package_agreements::
+        # the package name is appended to the end of this command, so we try here to make a command which conveys information only
+        package_add_command => "$(sys.winsysdir)\\WindowsPowerShell\\v1.0\\powershell.exe -Command \"Write-Host You must set some vars for package-method-winget to work;exit 1; rem Trying to add package:";
+}
+
+# switch to module specific namespace to avoid name collisions
+body file control
+{
+  namespace => "package_method_winget";
+}
+
+# package_method_winget bundle's purpose is to look at inputs/data for acceptance of source and package agreements
+# these MUST be agreed to in order for this package method to work properly.
+bundle agent package_method_winget
+{
+  classes:
+    "accept_source_agreements" expression => regcmp("^[yY][eE]?[sS]?", "${data:package_method_winget.accept_source_agreements}"),
+      scope => "namespace";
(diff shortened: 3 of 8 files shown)


management/package-module-winget/winget-installed.cf
```diff
@@ -0,0 +1,22 @@
+bundle agent winget_installed
+{
+  methods:
+    windows::
+      "powershell_execution_policy_set" usebundle => powershell_execution_policy_set("LocalMachine", "Unrestricted"),
```
craigcomstock, line 5: @nickanderson I think we talked about making this an opt-in class maybe? I can add a README explaining why it is needed, so that we can install winget in case it is needed to be installed which should only apply to older Windows versions so maybe "OK" if I change this to make the powershell execution policy set depend on `winget_installed` and `winget_module_installed` classes below. But... I think the module may always NOT be installed by default so would still be essentially "required" for this package module to operate properly.

I will be doing some testing on my Windows 10 and other newer server cloud instances now.

craigcomstock: updated.

management/package-module-winget/winget-installed.cf
```diff
@@ -0,0 +1,22 @@
...
+  commands:
+    windows.!winget_installed.execution_policy_ok::
+      "powershell.exe -File '${this.promise_dirname}/install-winget.ps1'"
+        contain => powershell;
+    windows.!winget_module_installed.execution_policy_ok::
+      "powershell.exe -File '${this.promise_dirname}/install-winget-powershell-module.ps1'"
+        contain => powershell;
+
+  methods:
+    windows.(winget_not_installed|winget_module_not_installed).data:winget_installed_no_powershell_execution_changes::
```
craigcomstock, line 19: To avoid "negative knowledge" I suppose an opt-in would be better:

```suggestion
    windows.(winget_not_installed|winget_module_not_installed).data:winget_installed_allow_powershell_execution_changes::
```

craigcomstock: @nickanderson opinion? If someone doesn't read the README and the policy they might get errors so maybe a report if these bits need installation by the ps1 scripts but the opt-in class is not defined?

management/package-module-winget/winget-installed.cf
```diff
@@ -0,0 +1,22 @@
...
+      "powershell_execution_policy_set" usebundle => powershell_execution_policy_set("LocalMachine", "Unrestricted"),
```
craigcomstock, line 20: Thinking about this change some more I wonder if we shouldn't rather sign our scripts: https://learn.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_signing?view=powershell-7.4

By default it seems AWS windows instances have LocalMachine set to RemoteSigned.

management/package-module-winget/README
```diff
@@ -0,0 +1,34 @@
+package-module-winget
```
nickanderson, line 1: I know that this is a build /module/ but the implementation here is a package /method/

So I would suggest renaming this to package-method-winget.

management/package-module-winget/README
```diff
@@ -0,0 +1,34 @@
...
+
+```cf3
+packages:
+  windows::
+    "Microsoft.WindowsTerminal"
+      package_method => winget,
+      package_policy => "add";
+```
+
+Note that this module uses both the `winget` command as well as the PowerShell WinGet Client cmdlets as they make it easier to gather software inventory of currently installed packages.
```
nickanderson, line 14:
```suggestion
Note that this module uses both the `winget` command as well as the `PowerShell WinGet Client` _cmdlets_ as they make it easier to gather software inventory of currently installed packages.
```

management/package-module-winget/README
```diff
@@ -0,0 +1,34 @@
...
+
+`winget` should be installed on most newer systems by default.
+
+In order for `winget` to be installed and Microsoft.WinGet.Client (the cmdlets module) ps1 scripts must be run which requires PowerShell Execution policy Unrestricted for LocalMachine.
```
nickanderson, line 18:
```suggestion
In order for `winget` to be installed and `Microsoft.WinGet.Client` (the cmdlets module) ps1 scripts must be run which requires `PowerShell Execution policy` `Unrestricted` for `LocalMachine`.
```

management/package-module-winget/README
```diff
@@ -0,0 +1,34 @@
...
+The policy by default will not make this change.
+You must opt-in by defining the class `winget_installed_allow_powershell_execution_changes` either in Groups/Host data or Augments in the namespace `data`:
+
+```json
+{
+  "classes": {
+    "data:winget_installed_allow_powershell_execution_changes": ["any"]
+  }
+}
+```
```
nickanderson, line 28: I would prefer the new more explicit (albeit complex style)

```json
{
  "classes": {
    "data:winget_installed_allow_powershell_execution_changes": {
      "class_expressions":  [ "windows::" ],
      "comment": "winget will not function without the ability to execute powershell scripts"
    }
  }
}
```

management/package-module-winget/package-module-winget.cf
```diff
@@ -0,0 +1,21 @@
...
+body package_method winget
+{
+      package_changes => "bulk";
+      package_name_convention   => "$(name)";
+      package_delete_convention => "$(name)";
+
+      package_installed_regex => ".*";
+      # Note that package_list_name_regex does not allow for commas inside of package names as parsing that from ConvertTo-Csv below would be too complex
+      package_list_name_regex => '^"([^"]*)",.*';
+      package_list_version_regex => '.*,"([^"]*)".*';
```
nickanderson, line 12: Would be nice to have some sample output comment stub to easily reference what these look to match against.

management/package-module-winget/package-module-winget.cf
```diff
@@ -0,0 +1,21 @@
...
+
+      # Here we use winget instead of PowerShell Cmdlets because we can provide the --accept-source-agreements and --accept-package-agreements this way which gets around dialog prompts
+      package_add_command    => "$(sys.winsysdir)\\WindowsPowerShell\\v1.0\\powershell.exe -Command \"winget install --accept-source-agreements --accept-package-agreements ";
```
nickanderson, line 15: I think this needs to be guarded against until the user has explicitly provided input to accept the agreements.

management/package-module-winget/winget-installed.cf
nickanderson: I am not sure we should acquire code signing certificates for this. 
Were you thinking of dynamically self signing the script on each host and then relying on that self signature to suffice?

cfbs.json
```diff
@@ -154,6 +154,25 @@
...
+    "powershell-execution-policy": {
+      "description": "Inventory and bundle for PowerShell Execution Policy",
+      "subdirectory": "management/powershell-execution-policy",
+      "steps": [
+        "directory ./ services/cfbs/powershell-execution-policy/",
+        "policy_files services/cfbs/powershell-execution-policy/",
+        "bundles powershell_execution_policy_inventory"
+      ]
+    },
+    "package-module-winget": {
```
nickanderson, line 166: I think this one should have module input for accepting the license agreements.

nickanderson (changes requested):

management/package-module-winget/winget-installed.cf
nickanderson: Yeah, good idea.

nickanderson: though, reports only goto console, they dont really come to mission portal sadly.

craigcomstock: yeah, I couldn't figure out or think of a way to get these bits into not kept promises that could be seen in policy analyzer/reports.

management/package-module-winget/winget-installed.cf
craigcomstock: Just thinking about what we should do other than "Unrestricted". I suppose self-signing scripts that we pull from a policy server makes "some sense" but is probably not the right answer. We can likely leave as-is.

management/package-module-winget/README
craigcomstock: changed, yes.

management/package-method-winget/package-method-winget.cf
```diff
@@ -0,0 +1,67 @@
+# switch to module specific namespace to avoid name collisions
+body file control
+{
+  namespace => "package_method_winget";
+}
+
+# package_method_winget bundle's purpose is to look at inputs/data for acceptance of source and package agreements
+# these MUST be agreed to in order for this package method to work properly.
+bundle agent package_method_winget
+{
```
nickanderson, line 10:
```suggestion
# @brief Verify user acceptance of source and package agreements
# these MUST be agreed to in order for this package method to work properly.
{
```

You can make that comment a brief description bit more in line with style of MPF. It's pretty low value though because the only tooling that uses that is our official docs site.

management/package-method-winget/package-method-winget.cf
```diff
@@ -0,0 +1,67 @@
...
+    # for group/host data, look in data:bundle.setting
+    # for cfbs inputs, look in bundle:bundle.setting as all three (namespace, bundle, variable) are required by cfbs
+    "var_accept_package_agreements" string => ifelse(
+        isvariable("data:package_method_winget.accept_package_agreements"),
+        "${data:package_method_winget.accept_package_agreements}",
+        "${package_method_winget:package_method_winget.accept_package_agreements}");
+    "var_accept_source_agreements" string => ifelse(
+        isvariable("data:package_method_winget.accept_source_agreements"),
+        "${data:package_method_winget.accept_source_agreements}",
+        "${package_method_winget:package_method_winget.accept_source_agreements}");
```
nickanderson, line 21: Sorry, it's not clear to me. Why do you have to look for these in 3 places? 

Why not just `package_method_winget:package_method_winget.accept_source_agreements` and `package_method_winget:package_method_winget.accept_package_agreements`?

management/package-method-winget/package-method-winget.cf
```diff
@@ -0,0 +1,67 @@
...
+
+  classes:
+    "accept_source_agreements" expression => regcmp("^[yY][eE]?[sS]?", "${var_accept_source_agreements}"),
+      scope => "namespace";
+    "accept_package_agreements" expression => regcmp("^[yY][eE]?[sS]?", "${var_accept_package_agreements}"),
+      scope => "namespace";
```
nickanderson, line 27: So accept in any case y, ye, yes

management/package-method-winget/package-method-winget.cf
craigcomstock: just two different namespaces: data (for host/group data) and bundle for inputs.

Ideally we should be registering these variables/classes with meta data that Mission Portal can scrape to auto-populate possible tunables in the UI but :cloud: (dreaming).

management/package-method-winget/README.md
```diff
@@ -0,0 +1,54 @@
...
+    info: Q:powershell.exe -Comm ...:You must set some vars for package-method-winget to work
+   error: Finished command related to promiser 'Microsoft.WindowsTerminal' -- an error occurred, returned 1
+   error: Bulk package schedule execution failed somewhere - unknown outcome for 'Microsoft.WindowsTerminal'
+```
+
+Acceptance can be given either via cfbs inputs, augments, group or host specific data.
+
+Set the value of `yes` in the following variables:
+- data:package_method_winget.accept_source_agreements
+- data:package_method_winget.accept_package_agreements
```
nickanderson, line 32: Is this documentation stale? I thought you circled into a single hinge for each of these but I see inputs targets the `package_method_winget` namespace instead of data as shown here.

Also, suggest verbatim for these.

```suggestion
- `data:package_method_winget.accept_source_agreements`
- `data:package_method_winget.accept_package_agreements`
```

nickanderson: Reading a bit further, it seems like maybe its the module input that hasn't been updated yet.

cfbs.json
```diff
@@ -154,6 +154,52 @@
...
+          "question": "Would you like to accept source agreements for winget inventory and packages promises? [yes|no]"
+        },
+        {
+          "type": "string",
+          "variable": "accept_package_agreements",
+          "namespace": "package_method_winget",
+          "bundle": "package_method_winget",
+          "label": "Accept Package Agreements",
+          "question": "Would you like to accept package agreements for winget inventory and packages promises? [yes|no]"
+        },
```
nickanderson, line 192: Didn't you move these to the data namespace?

nickanderson (changes requested):
I think the module input hasn't yet been updated to target the data namespace.
Minor suggestion of style for README mentioning variables (use verbatim)

Overall seems to make sense to me.

cfbs.json
craigcomstock: yes, fixed. :+1:

management/package-method-winget/README.md
craigcomstock: fixed

cfbs.json
```diff
@@ -154,6 +154,49 @@
...
+      "description": "Package method for Windows winget package manager.",
+      "subdirectory": "management/package-method-winget",
+      "dependencies": ["powershell-execution-policy"],
+      "steps": [
+        "input ./input.json def.json",
+        "directory ./ services/cfbs/modules/package-method-winget/",
+        "policy_files services/cfbs/modules/package-method-winget/",
+        "bundles package_method_winget:package_method_winget winget_installed:winget_installed"
+      ],
+      "input": [
```
nickanderson, line 176: I would explicitly target the `data` namespace here but that's pretty minor.

nickanderson:
Actually I just ran a test with cfbs, and it seems that namespace is required for input.

```
$ cfbs build
Error in cfbs.json for module 'lynis': The "namespace" field is required in module input elements
WARNING: At least one error encountered while validating your cfbs.json file.
Please see the error messages above and apply fixes accordingly.
If not fixed, these errors will cause your project to not build in future cfbs versions.

Modules:
001 masterfiles @ d5d5c61942a97ec5e3fd35478e590c5943db5ac5 (Downloaded)
002 lynis       @ 4db6d515c45873a9f3fa7ce78ece593027464b64 (Downloaded)

Steps:
001 masterfiles : run 'EXPLICIT_VERSION=3.24.0 EXPLICIT_RELEASE=1 ./prepare.sh -y'
001 masterfiles : copy './' 'masterfiles/'
002 lynis       : copy 'policy/main.cf' 'masterfiles/services/lynis/main.cf'
002 lynis       : json 'cfbs/def.json' 'masterfiles/def.json'
002 lynis       : input './input.json' 'masterfiles/def.json'

Generating tarball...

Build complete, ready to deploy 🐿
 -> Directory: out/masterfiles
 -> Tarball:   out/masterfiles.tgz

To install on this machine: sudo cfbs install
To deploy on remote hub(s): cf-remote deploy

```

snippet of my cfbs.json whre I removed the namespace key for the input step for setting tar_url.

```
      "input": [
        {
          "type": "string",
          "namespace": "lynis",
          "bundle": "globals",
          "variable": "version",
          "label": "Lynis version",
          "question": "What version of Lynis should be used?",
          "default": "3.1.1"
        },
        {
          "type": "string",
          "bundle": "globals",
          "variable": "tar_url",
          "label": "Tarball url",
          "question": "Where should clients download lynis from (http/https)",
          "default": "https://downloads.cisofy.com/lynis/lynis-$(version).tar.gz",
          "comment": "Some may want to self host the tarball within their own infrastructure. This provides an easy way to do that. Note, the archive should be named for the version of Lynis selected."
        },

```

nickanderson (changes requested):
I think you must set namespace for module input.

craigcomstock:
I didn't have to with cfbs 4.2.1 which is odd since we are both using that version. I suppose you are using a build index module and I am using a repo#commit. Maybe that's the diff.

nickanderson (approved):

PR #103, opened 2024-09-19
add newer class for purging policies

craigcomstock:
Running on 3.24.x I get the following report/warning

```
R: `cfengine_internal_purge_policies` no longer has any effect. Please use `cfengine_internal_purge_policies_disabled` instead, to choose where you want to disable purging or remove the class completely if you want purging enabled everywhere (the new default in 3.18+).
```

I figured we should add both? So that this demo module works on older versions but the note mentions 3.18+ so maybe we can just remove instead).

### DIFF
--- a/management/demo/def.json
+++ b/management/demo/def.json
@@ -2,6 +2,7 @@
     "classes": {
         "mpf_augments_control_enabled": ["any"],
         "cfengine_internal_purge_policies": ["any"],
+        "cfengine_internal_purge_policies_disabled": ["any"],
         "cfengine_mp_fr_dependencies_auto_install": ["any"]
     },
     "vars": {


management/demo/def.json
```diff
@@ -2,6 +2,7 @@
     "classes": {
         "mpf_augments_control_enabled": ["any"],
         "cfengine_internal_purge_policies": ["any"],
+        "cfengine_internal_purge_policies_disabled": ["any"],
```
nickanderson, line 5:
```suggestion
```

The class `cfengine_internal_purge_policies` would enable that feature on older versions of the MPF. When this feature was enabled files inside of /var/cfengine/inputs that did not exist on the servers /var/cfengine/masterfiles would be removed from the client, in effect a sync behavior instead of the default to just copy files on top if changes were made, leaving extra files alone.

That default changed and so now that behavior is the default. If you previously had  `cfengine_internal_purge_policies`  defined then today you should simply disable the definition of that class to keep that sync behavior. Alternatively, if you desire the old behavior where extra files were not cleaned up then you would instead define `cfengine_internal_purge_policies_disabled`.

Since, this was enabling that feature which is now the default, I think that both of these should just be removed.

nickanderson (changes requested):
I think we should simply delete both.

management/demo/def.json
nickanderson: After thinking about this for a moment another option would be to adjust `cfengine_internal_purge_policies` to target `cfengine_3_18` if the desire is to keep that behavior for older clients that may still be using this. Otherwise, still best to remove both i think.

nickanderson (changes requested):

management/demo/def.json
craigcomstock: agreed. I didn't read things quite right. Will adjust.

craigcomstock:
closed in preference for more appropriate change: https://github.com/cfengine/modules/pull/105

PR #107, opened 2024-10-23
Added setting of HOME environment variable for git promise type

craigcomstock:
There was an issue where due to the lack of the $HOME env var the git commands were not picking up my .gitconfig and .git-credentials files.

This change enables a way to setup authentication for private repositories.

.gitconfig could be:

[credential]
    helper = store

and .git-credentials could be something of the form:

https://<url-encoded-username>:<password>@gitlab.com

### DIFF
--- a/promise-types/git/git.py
+++ b/promise-types/git/git.py
@@ -1,5 +1,6 @@
 import os
 import subprocess
+from pathlib import Path
 
 from typing import Dict, List, Optional
 
@@ -255,6 +256,9 @@ class GitPromiseTypeModule(PromiseModule):
         env["GIT_SSH_COMMAND"] = model.ssh_executable
         if model.ssh_options:
             env["GIT_SSH_COMMAND"] += " " + model.ssh_options
+        if not 'HOME' in env:
+            # git should have a HOME env var to retrieve .gitconfig, .git-credentials, etc
+            env["HOME"] = str(Path.home())
         return env
 
 
--- a/promise-types/git/README.md
+++ b/promise-types/git/README.md
@@ -60,6 +60,33 @@ bundle agent main
 }
 ```
 
+## Authentication
+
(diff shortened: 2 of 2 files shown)


promise-types/git/git.py
```diff
@@ -236,6 +237,7 @@ def evaluate_promise(self, promiser: str, attributes: Dict):
 
     def _git(self, model: object, args: List[str], cwd: Optional[str] = None) -> str:
         self.log_verbose("Run: {cmd}".format(cmd=" ".join(args)))
+        os.environ['HOME'] = str(Path.home())
```
craigcomstock, line 240: would be better to check for HOME first, then set.

and while I'm at it, I think we should have an `as_user` attribute to run as a particular user.

promise-types/git/git.py
```diff
@@ -237,6 +238,10 @@ def evaluate_promise(self, promiser: str, attributes: Dict, metadata: Dict):
 
     def _git(self, model: object, args: List[str], cwd: Optional[str] = None) -> str:
         self.log_verbose("Run: {cmd}".format(cmd=" ".join(args)))
+        # In order for git to use .gitconfig, .git-credentials, etc HOME must be set.
+        # By default, cf-agent empties the environment and so we must set HOME if not already set.
+        if not 'HOME' in os.environ:
+            os.environ['HOME'] = str(Path.home())
```
vpodzime, line 244: Why not add it in `_git_envvars()` only for the `git` subprocess?

vpodzime (approved):
Probably fine either way.

larsewi (approved):
🚀

promise-types/git/git.py
craigcomstock: because that was unknown to me and not obvious? ;) It even has an underscore (secret).

But sure, makes sense. :)

craigcomstock: Sure, I see it now. Can do.

PR #113, opened 2025-06-24
CFE-4542: Added json promise type

victormlg:


promise-types/json/json_promise.py
```diff
@@ -0,0 +1,75 @@
...
+
+        if self.field:
+            content[self.field] = self.data
+        else:
+            content = self.data
+            
+        with open(self.filename, 'w') as f:
+            json.dump(content, f, indent=4)
+
+        return (Result.KEPT, ["written_json"])
```
olehermanse, line 71: Your promise type should result in promise kept if the file was already correct, and promise repaired if it was actually changed.

promise-types/json/json_promise.py
```diff
@@ -0,0 +1,75 @@
...
+        try:
+            with open(self.filename, 'r+') as f:
+                content = json.load(f)
+        except (FileNotFoundError, json.JSONDecodeError):
+            content = {}
+
+        if self.field:
+            content[self.field] = self.data
+        else:
+            content = self.data
```
olehermanse, line 66: Here you are relying on self.data set by `validate_promise()`. This is not compatible with what is written in the specification;

https://docs.cfengine.com/docs/3.24/reference-promise-types-custom.html#protocol-request-sequence

> Any promise must be validated before it is evaluated. When handling `evaluate_promise` requests, the module can assume that validation was already performed and successful, with the same attributes and promiser string. It is not guaranteed that the previous request was a `validate_promise` request. A `validate_promise` request does not imply that the next request will be `evaluate_promise`.

In other words, according to the spec, the agent could choose to send 5 different validate requests, and then 5 different evaluate requests, and your code would not work correctly.

promise-types/json/json_promise.py
```diff
@@ -0,0 +1,75 @@
...
+import os
+
+class JsonPromiseTypeModule(PromiseModule):
+
+
+    def __init__(self, **kwargs):
+        super(JsonPromiseTypeModule, self).__init__(
+            name="json_promise_module",
+            version="0.0.1",
+            record_file_path="/tmp/OUTPUT.txt"
```
olehermanse, line 13: Remove or leave a comment about what this is for.

promise-types/json/json_promise.py
```diff
@@ -0,0 +1,75 @@
...
+        self.add_attribute("array", list)
+        self.add_attribute("string", str)
+        self.add_attribute("true", str)
+        self.add_attribute("false", str)
+        self.add_attribute("null", str)
+
+
+    def validate_promise(self, promiser, attributes, metadata):
+
+        self.filename, _, self.field = promiser.partition(":")
```
olehermanse, line 27:
```suggestion
        filename, _, field = promiser.partition(":")
```

promise-types/json/json_promise.py
```diff
@@ -0,0 +1,75 @@
...
+
+        if os.path.exists(self.filename) and not os.path.isfile(self.filename):
+            raise ValidationError("'{}' already exists and is not a file".format(self.filename))
+
+        if not self.filename.endswith(".json"):
+            raise ValidationError("'{}' has to be a json file")
```
olehermanse, line 33:
```suggestion
        if os.path.exists(filename) and not os.path.isfile(filename):
            raise ValidationError("'{}' already exists and is not a file".format(filename))

        if not filename.endswith(".json"):
            raise ValidationError("'{}' has to be a json file".format(filename))
```

promise-types/json/json_promise.py
```diff
@@ -0,0 +1,75 @@
...
+        except (FileNotFoundError, json.JSONDecodeError):
+            content = {}
+
+        if self.field:
+            content[self.field] = self.data
+        else:
+            content = self.data
+            
+        with open(self.filename, 'w') as f:
+            json.dump(content, f, indent=4)
```
olehermanse, line 69: Writing should only happen if necessary. And when it is happening, there should be an `INFO` level log message.

promise-types/json/json_promise.py
```diff
@@ -0,0 +1,75 @@
+from cfengine import PromiseModule, ValidationError, Result
+from typing import Any
+import json
+import os
```
olehermanse, line 4: Very small nitpick, but it's relatively common to order / group your imports, so standard library is first, etc.

```suggestion
import os
import json
from typing import Any

from cfengine import PromiseModule, ValidationError, Result
```

promise-types/json/json_promise_type.py
```diff
@@ -0,0 +1,152 @@
...
+        else:
+            if content == data:
+                return (Result.KEPT, ["correct_json"])
+            content = data
+
+        with open(filename, "w") as f:
+            json.dump(content, f, indent=4)
+
+        self.log_info("Updated '{}'".format(filename))
+        return (Result.REPAIRED, ["written_json"])
```
olehermanse, line 148: As mentioned in Slack - I don't think it's necessary to use the result classes feature in your module, it doesn't provide any additional information. If users want to define classes based on the outcome, they can use a classes body.

```suggestion
        return Result.REPAIRED
```

promise-types/json/json_promise_type.py
```diff
@@ -0,0 +1,152 @@
...
+        except (FileNotFoundError, json.JSONDecodeError):
+            content = {}
+
+        if field:
+            if field in content and content[field] == data:
+                return (Result.KEPT, ["correct_json"])
+            content[field] = data
+        else:
+            if content == data:
+                return (Result.KEPT, ["correct_json"])
```
olehermanse, line 141:
```suggestion
                return Result.KEPT
```

promise-types/json/json_promise_type.py
```diff
@@ -0,0 +1,152 @@
...
+
+        try:
+            with open(filename, "r+") as f:
+                content = json.load(f)
+        except (FileNotFoundError, json.JSONDecodeError):
+            content = {}
+
+        if field:
+            if field in content and content[field] == data:
+                return (Result.KEPT, ["correct_json"])
```
olehermanse, line 137:
```suggestion
                return Result.KEPT
```

promise-types/json/json_promise_type.py
```diff
@@ -0,0 +1,152 @@
...
+                raise ValidationError(
+                    "'{}' is not a valid data array".format(model.array)
+                )
+
+        if model.number and not is_number(model.number):
+            raise ValidationError(
+                "'{}' is not a valid int or real".format(model.number)
+            )
+
+        if model.primitive and model.primitive not in ["true", "false", "null"]:
```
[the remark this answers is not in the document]
olehermanse: @larsewi Not according to the json spec; https://www.json.org/json-en.html

The problem is that we need to have a word (an attribute) for `true or false or null`. We chose primitive as the best option, and there will be clear information in both documentation and warnings about what it means.

promise-types/json/json_promise_type.py
```diff
@@ -0,0 +1,152 @@
...
+                "The promiser '{}' is missing a type attribute. The possible types are {}".format(
+                    promiser, str(self.types)
+                )
+            )
+        elif len(present_types) > 1:
+            raise ValidationError(
+                "The attributes {} cannot be together".format(str(self.types))
+            )
+
+        filename, _, _ = promiser.partition(":")
```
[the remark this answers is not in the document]
olehermanse: I think this is an entirely reasonable limitation to have, especially for a new experimental module.

olehermanse (approved):
I don't see anything significant blocking this PR - I think you can review and consider / address all the comments in a smaller followup PR.

PR #114, opened 2025-06-30
Fixed json promise type

victormlg:
Follow-up PR to the one introducing json promise type

### DIFF
--- a/promise-types/json/json_promise_type.py
+++ b/promise-types/json/json_promise_type.py
@@ -1,5 +1,7 @@
 import os
 import json
+import tempfile
+import shutil
 
 from cfengine import PromiseModule, ValidationError, Result, AttributeObject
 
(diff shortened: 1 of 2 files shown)


promise-types/json/json_promise_type.py
```diff
@@ -133,19 +133,32 @@ def evaluate_promise(self, promiser, attributes, metadata):
...
         else:
             if content == data:
-                Result.KEPT
+                return Result.KEPT
             content = data
 
-        with open(filename, "w") as f:
+        fd, tmp = tempfile.mkstemp()
+
+        with open(tmp, "w") as f:
```
larsewi, line 154: > [mkstemp()](https://docs.python.org/3/library/tempfile.html#tempfile.mkstemp) returns a tuple containing an OS-level handle to an open file (as would be returned by [os.open()](https://docs.python.org/3/library/os.html#os.open)) and the absolute pathname of that file, in that order.

So you don't need to open it again

promise-types/json/json_promise_type.py
```diff
@@ -133,19 +133,32 @@ def evaluate_promise(self, promiser, attributes, metadata):
...
+                return Result.KEPT
             content[field] = data
         else:
             if content == data:
-                Result.KEPT
+                return Result.KEPT
             content = data
 
-        with open(filename, "w") as f:
+        fd, tmp = tempfile.mkstemp()
```
larsewi, line 152: I think you can use `mkstemp()` with the `with` keyword. And you don't have to worry about closing it

promise-types/json/json_promise_type.py
```diff
@@ -133,19 +133,32 @@ def evaluate_promise(self, promiser, attributes, metadata):
...
+
+        with open(tmp, "w") as f:
             json.dump(content, f, indent=4)
 
+        shutil.move(tmp, filename)
+        os.close(fd)
```
larsewi, line 158: Maybe close file descriptor before moving the file

promise-types/json/json_promise_type.py
```diff
@@ -106,23 +111,18 @@ def evaluate_promise(self, promiser, attributes, metadata):
         model = self.create_attribute_object(attributes)
         filename, _, field = promiser.partition(":")
 
+        if os.path.exists(filename) and not os.path.isfile(filename):
+            self.log_error("'{}' already exists and is not a file".format(filename))
```
larsewi, line 115:
```suggestion
            self.log_error("'{}' already exists and is not a regular file".format(filename))
```

promise-types/json/json_promise_type.py
```diff
@@ -133,19 +133,36 @@ def evaluate_promise(self, promiser, attributes, metadata):
             content = {}
```
larsewi, line 133: You should add
```Python3
except Exception as e:
    self.log_error("Failed to read '{}': {}".format(filename, e))
    return Result.NOT_KEPT
```
to catch all other exceptions raised by `open()` to avoid the module from crashing with a stack trace and instead return promise not kept.

PR #128, opened 2026-03-12
Custom promise type for managing global sshd configuration

larsewi:


promise-types/sshd/sshd.py
```diff
@@ -0,0 +1,570 @@
...
+        i += 1
+    return directive
+
+
+def normalize_directive(directive: str) -> str:
+    """Normalize a directive by removing trailing comments and replacing = with a space."""
+    directive = strip_trailing_comment(directive)
+    # Normalize separator (= or whitespace) to a single space
+    directive = re.sub(r"\s*=\s*", " ", directive, count=1)
+    return directive.strip().lower()
```
olehermanse: Docstring could mention why / when this is used.

olehermanse (approved):
I think it's good enough for an experimental / first iteration, but there's many things we could consider doing differently. Left some comments in the code.

olehermanse (approved):
Looks good. Some small comments / suggestions.

PR #133, opened 2026-04-03
Fixes and refactor for appstreams promise type

nickanderson:


larsewi (approved):
For future PRs, if possible, try to split up changes into multiple commits and add some context as to what you are fixing.

It would also be nice if you added references in the commit messages to support claims like:

> DNF's module database can mark a profile as installed even if the RPM transaction failed (e.g. due to a stale cache error)

nickanderson:
> For future PRs, if possible, try to split up changes into multiple commits and add some context as to what you are fixing.

Fair

> It would also be nice if you added references in the commit messages to support claims like:
> 
> > DNF's module database can mark a profile as installed even if the RPM transaction failed (e.g. due to a stale cache error)

You mean like cli output that showed the weird state on the system from my testing or you mean links to some other documentation?

olehermanse:
> You mean like cli output that showed the weird state on the system from my testing or you mean links to some other documentation?

@nickanderson one, or the other, or both :)

PR #141, opened 2026-05-06
CFE-4657: Set up GH Action which tests old python in cfengine/modules

sarakthon:
Also fix composite action to add pyenv to PATH, since pyenv-action step is skipped on cache hit

Ticket: CFE-4657

### DIFF
--- a/.github/workflows/python-tests.yml
+++ b/.github/workflows/python-tests.yml
@@ -1,0 +1,26 @@
+name: Python tests
+
+on:
+  push:
+    branches: [master]
+  pull_request:
+    branches: [master]
+
+jobs:
+  test-legacy:
+    runs-on: ubuntu-24.04
+    permissions:
+      contents: read
+    strategy:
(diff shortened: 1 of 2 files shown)


.github/workflows/python-tests.yml
```diff
@@ -0,0 +1,26 @@
...
+
+jobs:
+  test-legacy:
+    runs-on: ubuntu-24.04
+    permissions:
+      contents: read
+    strategy:
+      fail-fast: false
+      matrix:
+        python-version: ["3.5.10", "3.6.15", "3.7.10"]
```
larsewi: `cfbs` also tests `["3.8", "3.9", "3.10", "3.11", "3.12", "3.13"]`. We should probably do the same here?

sarakthon: I have now added these versions as well

larsewi (approved):

PR #143, opened 2026-05-13
Added dirtyfrag module

nickanderson:


security/dirtyfrag/dirtyfrag.cf
```diff
@@ -0,0 +1,402 @@
...
+    # --- Constants ---
+    "_esp_conf_path" string => "/etc/modprobe.d/dirtyfrag-esp.conf";
+    "_rxrpc_conf_path" string => "/etc/modprobe.d/dirtyfrag-rxrpc.conf";
+
+    "_esp_conf_content"
+      string => concat(
+        "# Dirty Frag CVE-2026-43284 mitigation: block xfrm-ESP and IPComp$(const.n)",
+        "install esp4 /bin/false$(const.n)",
+        "install esp6 /bin/false$(const.n)",
+        "install ipcomp4 /bin/false$(const.n)",
```
oldgiova: Maybe you have to drop the `4`? 

```
"install ipcomp /bin/false$(const.n)",
```

security/dirtyfrag/README.md
```diff
@@ -0,0 +1,138 @@
...
+
+Each CVE has an independent toggle and separate conf file:
+
+**CVE-2026-43284** (ESP/IPComp) -- `/etc/modprobe.d/dirtyfrag-esp.conf`:
+
+```
+# Dirty Frag CVE-2026-43284 mitigation: block xfrm-ESP and IPComp
+install esp4 /bin/false
+install esp6 /bin/false
+install ipcomp4 /bin/false
```
oldgiova: Same here: maybe you have to drop the `4`?

security/dirtyfrag/dirtyfrag.cf
```diff
@@ -0,0 +1,402 @@
...
+    "_rxrpc_conf_exists" expression => fileexists("${_rxrpc_conf_path}");
+    "_userns_conf_exists" expression => fileexists("${_userns_conf_path}");
+
+    # ESP is mitigated by modprobe blacklist, disabled userns, patched kernel,
+    # or admin override
+    "_esp_mitigated"
+      or => {
+        "_esp_conf_exists",
+        "_userns_conf_exists",
+        "!_userns_enabled",
```
oldgiova: RedHat/SUSE have `/proc/sys/kernel/unprivileged_userns_clone`, so `!_userns_enabled` is always true, so `_esp_mitigated` is true. The consequence is that every RHEL/SLES host with esp4.ko on disk are listed as mitigated always.

Maybe probe `user.max_user_namespaces` instead as an evidence of mititagion?

security/dirtyfrag/dirtyfrag.cf
```diff
@@ -0,0 +1,402 @@
...
+
+    "_ipcomp6_path"
+      string => "/lib/modules/$(_kver)/kernel/net/ipv6/ipcomp6.ko";
+
+    "_rxrpc_path" string => "/lib/modules/$(_kver)/kernel/net/rxrpc/rxrpc.ko";
+    "_ns_proc" string => "/proc/sys/kernel/unprivileged_userns_clone";
+
+    # --- Read the unprivileged user namespace setting ---
+    "_ns_val"
+      string => readfile("${_ns_proc}"),
```
oldgiova: readfile() does not strip trailing newlines, so the comparison is false even if the file exists. Maybe use `sysctlvalue()` ?

```
  vars:
    "_max_userns" string => sysctlvalue("user.max_user_namespaces");

  classes:
    "_userns_disabled" expression => strcmp("0", "${_max_userns}");
```

Then in _esp_mitigated, drop the userns leg entirely, and fold it into the vulnerability condition instead:
```
  "dirtyfrag_esp_needs_mitigation"
    and => { "dirtyfrag_esp_present", "!_userns_disabled", "!_esp_mitigated" };
```

security/dirtyfrag/dirtyfrag.cf
```diff
@@ -0,0 +1,402 @@
...
+  classes:
+    "_patch_data_matched"
+      expression => isvariable("_matched_idx");
+
+    # Use sort -V to check: if the patched version sorts <= running version,
+    # then the running kernel is patched. We test by checking that the
+    # patched version comes first (or equal) in version-sorted order.
+    # printf '%s\n' "$patched" "$running" | sort -V | head -1
+    # If result == patched, then running >= patched.
+    "_esp_kernel_patched"
```
oldgiova: These classes are not passed by default in main, so you should widen the scope, right?

security/dirtyfrag/patched-kernels.json
```diff
@@ -0,0 +1,66 @@
...
+      "https://almalinux.org/blog/2026-05-07-dirty-frag/",
+      "https://www.suse.com/security/cve/CVE-2026-43284.html",
+      "https://explore.alas.aws.amazon.com/CVE-2026-43284.html"
+    ]
+  },
+  "entries": [
+    {
+      "label": "RHEL/CentOS/Alma/Rocky/OL 8",
+      "id_match": "^(rhel|centos|rocky|almalinux|ol)$",
+      "version_match": "^8",
```
oldgiova: This doesn't match `8.10` is it?

oldgiova (commented):
I've asked Claude Code to helping me review this and here's the finding (the ones that I understood and confirmed with human double-checks)

security/dirtyfrag/README.md
```diff
@@ -0,0 +1,138 @@
...
+
+- **Dirty Frag CVE-2026-43284 (xfrm-ESP) status**:
+  - `VULNERABLE (esp4, esp6 loaded)` -- vulnerable modules currently in memory (names vary by host)
+  - `VULNERABLE (modules on disk, none loaded)` -- modules present but not loaded; latent risk
+  - `PATCHED (kernel fix applied)` -- running kernel version includes the fix (auto-detected or admin-declared)
+  - `MITIGATED (blacklist in place)` -- modprobe blacklist or userns restriction active
+  - `NOT AFFECTED` -- vulnerable modules not present on this host
+- **Dirty Frag CVE-2026-43500 (RxRPC) status**:
+  - `VULNERABLE (rxrpc loaded)` -- module currently in memory
+  - `VULNERABLE (module on disk, not loaded)` -- module present but not loaded; latent risk
```
craigcomstock, line 26: this is the same as the above, should have some differentiation to know which CVE is "module on disk, not loaded".

security/dirtyfrag/README.md
```diff
@@ -0,0 +1,138 @@
...
+  - `PATCHED (kernel fix applied)` -- running kernel version includes the fix (auto-detected or admin-declared)
```
craigcomstock, line 27: ah, maybe the inventory name is the name of the CVE and the values what you are enumerating here... in that case the values should NOT include any mention of the CVE and maybe even not include comments so that they are more easily queried for just the CAPS portion.

security/dirtyfrag/README.md
```diff
@@ -0,0 +1,138 @@
...
+```
+# Dirty Frag CVE-2026-43284 mitigation: disable unprivileged user namespaces
+# Blocks ESP/IPComp exploit without disabling IPsec.
+# WARNING: May affect rootless containers and sandboxed applications.
+user.max_user_namespaces = 0
+```
+
+This blocks the ESP/IPComp exploit path without blacklisting the modules, preserving IPsec functionality. Use this instead of `mitigate_esp` on hosts that require IPsec. Note: this does **not** mitigate CVE-2026-43500 (RxRPC) and may break rootless containers (Podman, Docker rootless), Flatpak, and browser sandboxes. Applied via `sysctl --system` on first write.
+
+All mitigations are **disabled by default** -- the module only reports status unless the corresponding CMDB variable is set to `"true"`.
```
craigcomstock: what is the CMDB variable name?

security/dirtyfrag/dirtyfrag.cf
```diff
@@ -0,0 +1,402 @@
...
+      expression => returnszero(
+        "/usr/bin/test \"$(const.dollar)(/usr/bin/printf '%s\n' '${_esp_patched_ver}' '$(default:sys.release)' | /usr/bin/sort -V | /usr/bin/head -1)\" = '${_esp_patched_ver}'",
```
craigcomstock: maybe use paths here to avoid trouble with OS different paths to these common commands? You are already specifying useshell so might as well remove the full paths to the commands for flexibility sake.

security/dirtyfrag/dirtyfrag.cf
```diff
@@ -0,0 +1,402 @@
...
+        "useshell"
+      ),
+      if => isvariable("_esp_patched_ver");
+
+    "_rxrpc_kernel_patched"
+      expression => returnszero(
+        "/usr/bin/test \"$(const.dollar)(/usr/bin/printf '%s\n' '${_rxrpc_patched_ver}' '$(default:sys.release)' | /usr/bin/sort -V | /usr/bin/head -1)\" = '${_rxrpc_patched_ver}'",
```
craigcomstock: remove full paths to executables, useshell is in effect.

security/dirtyfrag/dirtyfrag.cf
```diff
@@ -0,0 +1,402 @@
...
+# CMDB variables (set via def.json augments):
+#   {
+#     "variables": {
+#       "dirtyfrag:main.mitigate_esp":      { "value": "true" },
+#       "dirtyfrag:main.mitigate_rxrpc":    { "value": "true" },
+#       "dirtyfrag:main.mitigate_userns":   { "value": "true" },
+#       "dirtyfrag:main.esp_patched":       { "value": "true" },
+#       "dirtyfrag:main.rxrpc_patched":     { "value": "true" }
+#     }
+#   }
```
craigcomstock: Here's where I think you should really REMOVE these bits from the README and instead refer to this comment here or... remove these in the policy and refer to the README. do not repeat yourself right? ;)

security/dirtyfrag/dirtyfrag.cf
```diff
@@ -0,0 +1,402 @@
...
+# not yet in the data file, or custom/backported kernels.
+#
+# When disabled (default), the module only reports vulnerability status.
+{
+  methods:
+    # Run kernel patch check before evaluating status
+    "kernel_patch_check" usebundle => "dirtyfrag:kernel_patch_check";
+
+  vars:
+    # --- Constants ---
```
craigcomstock: probably remove this comment, these are not constants in language terms though they may be in effect such things in usage. 99% of vars in cfengine policy language are vars right? even though you can redefine them at will.

security/dirtyfrag/dirtyfrag.cf
```diff
@@ -0,0 +1,402 @@
...
+        "dirtyfrag_rxrpc_needs_mitigation._rxrpc_loaded",
+        "VULNERABLE (rxrpc loaded)",
+        "dirtyfrag_rxrpc_needs_mitigation",
+        "VULNERABLE (module on disk, not loaded)",
+        "dirtyfrag:kernel_patch_check._rxrpc_kernel_patched|_rxrpc_admin_patched",
+        "PATCHED (kernel fix applied)",
+        "dirtyfrag_rxrpc_present._rxrpc_mitigated",
+        "MITIGATED (blacklist in place)",
+        "NOT AFFECTED"
+      );
```
craigcomstock, line 175: for all of these I mentioned in the README that maybe simpler values would be easier to see and query and the inventory NAME will convey which CVE is in play for that value so no need to repeat it in the value.

security/dirtyfrag/dirtyfrag.cf
```diff
@@ -0,0 +1,402 @@
...
+      string => "CVE-2026-43500 (RxRPC): $(_rxrpc_status)",
+      meta => {
+        "inventory",
+        "attribute_name=Dirty Frag CVE-2026-43500 (RxRPC) status",
+      },
+      comment => "CVE-2026-43500 RxRPC mitigation status";
+
+  classes:
+    # --- CMDB toggles ---
+    "_mitigate_esp" expression => strcmp("true", "$(mitigate_esp)");
```
craigcomstock: why vars with value "true" instead of classes in CMDB? It would seem that classes would make more sense. Maybe add a note here that you use variables instead of classes because you want to enable global settings like groups for all devices and then override individual hosts with CMDB values other than true. We can't specify CMDB data that REMOVES a class so this workaround is needed. Right?

security/dirtyfrag/dirtyfrag.cf
```diff
@@ -0,0 +1,402 @@
...
+    "_userns_enabled" and => { "_ns_proc_file_exists", "_ns_val_is_1" };
+
+    # --- xfrm-ESP/IPComp modules present? ---
+    "_esp4_on_disk" expression => fileexists("${_esp4_path}");
+    "_esp6_on_disk" expression => fileexists("${_esp6_path}");
+    "_ipcomp4_on_disk" expression => fileexists("${_ipcomp4_path}");
+    "_ipcomp6_on_disk" expression => fileexists("${_ipcomp6_path}");
+
+    # Compressed variants (.ko.zst, .ko.xz)
+    "_esp4_on_disk_z" expression => fileexists("${_esp4_path}.zst");
```
craigcomstock: I suppose you could have iterated on these bits, supplying an slist of suffixes of [ .ko, .ko.zst, .ko.xz ]

security/dirtyfrag/dirtyfrag.cf
```diff
@@ -0,0 +1,402 @@
...
+    # --- CVE-2026-43284 alternative: disable unprivileged user namespaces ---
+    _mitigate_userns::
+      "${_userns_conf_path}"
+        create => "true",
+        content => "${_userns_conf_content}",
+        classes => default:results("bundle", "dirtyfrag_userns_conf");
+
+  commands:
+    # --- Unload ESP/IPComp modules after conf written ---
+    _mitigate_esp.dirtyfrag_esp_conf_repaired::
```
craigcomstock: are you going to miss things sometimes if only when _repaired? Maybe a kernel module promise would be better... promise that these are not loaded.

craigcomstock (commented):
looks good. :)

nickanderson:
Thanks for the reviews, great stuff. I am working on an overhaul

security/dirtyfrag/dirtyfrag.cf
nickanderson: Yeah, it was at least concern about shadowing and not overriding classes to false that drove me towards a variable here.

security/dirtyfrag/dirtyfrag.cf
nickanderson: Yeah, could have.

security/dirtyfrag/dirtyfrag.cf
nickanderson: I don't love the overhead of custom promise types. semantically reading is way nicer, but it's a bit of yeuck for me every time i think about custom promises that would be backed with an interpreter.

but yes, keying on just repaired can miss stuff. Better to not key off indirect state and keyu off fact that module is loaded and mitigation is desired.

security/dirtyfrag/dirtyfrag.cf
nickanderson: :thinking:

nickanderson:
OK, fixups from reviews See what you think now

craigcomstock (approved):

PR #149, opened 2026-07-16
Added policy-dispatcher experimental module

olehermanse:


management/policy-dispatcher/main.cf
```diff
@@ -0,0 +1,79 @@
...
+{
+  classes:
+    "enabled" expression => isvariable("policy_files_to_run");
+    "run_$(i)" expression => "$(_condition[$(i)])";
+
+  vars:
+    enabled::
+      "i" slist => getindices(policy_files_to_run);
+
+      "_policy[$(i)]"
```
nickanderson: The policy content can be quite long and contain all kinds of special characters. I think using the content as an index could be problematic.

management/policy-dispatcher/main.cf
```diff
@@ -0,0 +1,79 @@
...
+        );
+
+      "_ifelapsed[$(i)]"
+        string => ifelse(
+          not(strcmp("$(policy_files_to_run[$(i)][ifelapsed])", "")),
+          "$(policy_files_to_run[$(i)][ifelapsed])",
+          "5"
+        );
+
+      "_file[$(i)]" string => "$(sys.statedir)/policy-dispatcher-$(i).cf";
```
nickanderson: Maybe change the policy-dispatcher- prefix into a directory policy-dispatcher/

management/policy-dispatcher/main.cf
```diff
@@ -0,0 +1,79 @@
...
+  reports:
+    enabled::
+      "Policy [$(i)]: $(_file[$(i)]), condition: $(_condition[$(i)]), ifelapsed: $(_ifelapsed[$(i)])";
+
+    !enabled::
+      "Policy-dispatcher: policy_files_to_run variable not found";
+
+  commands:
+    enabled::
+      "$(sys.cf_agent) --no-lock --file $(_file[$(i)])"
```
nickanderson, line 62: Oh, these are being run as full policy entries.
Makes me concerned about not breaking reporting with concurrent agent runs.

nickanderson (commented):
Interesting.

management/policy-dispatcher/main.cf
olehermanse: Yea, experimental.

management/policy-dispatcher/main.cf
olehermanse: They are numbers, not contents.

craigcomstock (approved):
looks interesting. I wonder if the README should include a bit more of what problem(s) it tries to solve and what advantage there is in using this module.

management/policy-dispatcher/main.cf
olehermanse: Done

larsewi (approved):

management/policy-dispatcher/main.cf
larsewi: As long as you don't run the "sub" agent in the background (i.e., you wait for it to exit), it should not mess with reporting.

PR #153, opened 2026-07-30
ENT-14326: Various fixes to the Ansible promise type

larsewi:
- **Narrow ansible ModuleNotFoundError guard to the imports**
- **Install ansible callback so it survives ansible-core 2.19**
- **Read the ansible playbook attribute from the attribute model**
- **Apply the ansible limit attribute**
- **Stop reporting skipped ansible tasks as errors**

### DIFF
--- a/promise-types/ansible/README.md
+++ b/promise-types/ansible/README.md
@@ -23,7 +23,7 @@ bundle agent main
 | ------------------ | --------- | ------------------------------------------------------------ | --------- | --------------- |
 | `playbook`         | `string`  | Absolute path of the Ansible playbook                        | No        | Promiser        |
 | `inventory`        | `string`  | Absolute path of the inventory file                          | No        | -               |
-| `limit`            | `slist`   | List of host names to target                                 | No        | `{"localhost"}` |
+| `limit`            | `slist`   | List of hosts and groups to target                           | No        | `{}` (no limit) |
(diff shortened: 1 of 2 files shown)


promise-types/ansible/README.md
```diff
@@ -23,7 +23,7 @@ bundle agent main
 | ------------------ | --------- | ------------------------------------------------------------ | --------- | --------------- |
 | `playbook`         | `string`  | Absolute path of the Ansible playbook                        | No        | Promiser        |
 | `inventory`        | `string`  | Absolute path of the inventory file                          | No        | -               |
-| `limit`            | `slist`   | List of host names to target                                 | No        | `{"localhost"}` |
+| `limit`            | `slist`   | List of hosts and groups to target                           | No        | `{}` (no limit) |
```
olehermanse, line 26: I think the old documented / intended default makes more sense for us (run on only localhost by default) even if it never actually worked as documented.

olehermanse (approved):

promise-types/ansible/README.md
larsewi: Yeah, I was just thinking, since it has been working as if it was no-limit -- due to the fact that it was never implemented -- then this would be the backwards compatible approach.

victormlg (approved):

